PR #4, opened 2020-10-19
Update build_extra_plugins.sh for lv2-dev

micahvdm:
lv2-dev needs to be added in order to successfully install on a clean patchbox os


rreichenbach (approved):
When using the top level setup.sh script, lv2-dev will get installed by pi-stomp/setup/pkgs/lilv_install.sh before build_extra_plugins.sh is called.  But if this script is run stand-alone lv2-dev won't be there so yeah, this certainly can't hurt and does help the stand-alone install case.

PR #14, opened 2021-09-20
Raspberry Pi Bullseye upstream fix

micahvdm:
Bullseye upstream is causing apt-get update to fail even with --fix-missing, so --allow-releaseinfo-change needs to be added to get around this.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -43,7 +43,7 @@ else
 fi
 
 printf "\n===== OS update =====\n"
-sudo apt update -y --fix-missing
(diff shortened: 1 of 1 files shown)


rreichenbach (commented):
That's probably fine, but we still need the -y so that it'll run without user approval

micahvdm:
Oh man, missed that I removed the -y, it is definitely necessary.

rreichenbach:
Accepting, then I'll add by the -y

PR #15, opened 2021-10-26
Setup fix and interrupt based encoder

ozbenh:
This has a small "fix" for the setup script so that it will properly configure the audio card if the user has gone
through the patchbox initial setup wizard (which tries to force you to select one of the two built-in audio
devices, hdmi and headphones).

There's also a change to the encoder code to use interrupt driven GPIOs. This makes my encoder works, it's otherwise unusable (unless you turn it very very slowly as to increase the duration of the "pulses"). I am pretty confident a lot of the problems people have been experiencing are related to encoders generating too fast pulses with a too low sampling rate.

I've run a number of tests with a pedal board hovering around 80% idle, with and without interrupts, and I seem to
get about the same amount of small overruns when actively manipulating the encoder. (a dozen after 2 or 3 seconds of manipulating, more if I use the "settings" menu). I get no overruns with a pedal board idling at 73%. In any case, I don't seem to see a negative impact to using interrupts.

The code is a bit refactored from what I posted to the forums: a use_interrupt argument to the constructor (default to True) was added so it's easy to revert to the old behaviour if a problem happens (at the very least as a way to test if this is causing said problem).

### DIFF
--- a/pistomp/encoder.py
+++ b/pistomp/encoder.py
@@ -14,34 +14,14 @@
 # along with pi-stomp.  If not, see <https://www.gnu.org/licenses/>.
 
 import RPi.GPIO as GPIO
+import threading
 
 from functools import partial
 
 
 class Encoder:
 
-    def __init__(self, d_pin, clk_pin, callback):
-
-        self.d_pin = d_pin
-        self.clk_pin = clk_pin
-        self.callback = callback
-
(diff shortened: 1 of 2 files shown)


rreichenbach:
This looks good.  I've tested on multiple pi-Stomp v1 and Core builds.  Fast rotations are tracked better on the LCD.

PR #16, opened 2021-11-01
Footswitch halo behaviour changes

ozbenh:
This makes footswitches that aren't bound to a MIDI message or relay
not toggle. It also synchronizes the LCD halo with the LED one if any,
so that the LCD halo of the relay switch toggles along with the relay
as happens on "normal" pedals.


rreichenbach (approved):
Looks good and tests well on v1 and Core

PR #17, opened 2021-11-01
Audiocard fixes

ozbenh:
Small fixes to the audiocard code, useful for the test backend


rreichenbach:
It works with either Capture Volume or Capture, so Ok.

PR #18, opened 2021-11-01
Switches GPIO fixes

ozbenh:
[Note: This branch is on top of the halo behaviour change one, I can rebase if you merge the previous one first]

This series reworks the footswitch and encoder switch code to avoid calling back into the core of pistomp from
the GPIO callbacks. Those callbacks are called by separate threads created by the GPIO library and we lack
the necessary synchronization to make it safe.

While at it, move the logic to a common subclass to both type of switch


ozbenh:
Rebased on top of master

rreichenbach:
Seems to work well

PR #20, opened 2021-11-01
Test backend

ozbenh:
This adds a test host type which can be used to validate the hardware in an ssh console before soldering the LCD

Note: the first commit is also in the footswitch-halo branch, but git should sort it out fine. It's recommended to at
least merge "Audiocard fixes" to avoid nasty stderr messages clobbering the curses display


ozbenh:
Rebased on top of master

rreichenbach:
This is super cool and useful.

PR #23, opened 2021-11-05
[RFC] wifi: Rework/improve wifi status

ozbenh:
This removes reliance on the patchbox CLI which is very slow and was only
called occasionally.

Instead, we have our own implementation to query the wifi state more
directly (much faster), running from a separate thread as to avoid
affecting the UI responsiveness.

This allows us to keep a "live" status that is properly updated when
things change from outside of pi-stomp control. This also cuts a few
seconds from pi-stomp startup time (patchbox CLI is that slow).

The next step is probably to have our own UI for configuring the wifi
but that is for another day :-) We probably want to do some serious
UI cleanup work before we do any drastic change such as implementing
an encoder-driven keyboard...

Signed-off-by: Benjamin Herrenschmidt <benh@kernel.crashing.org>

### DIFF
--- a/modalapi/mod.py
+++ b/modalapi/mod.py
@@ -86,7 +86,7 @@ class Mod(Handler):
     __single = None
 
     def __init__(self, audiocard, homedir):
-        self.wifi_monitor = None
+        self.wifi_manager = None
 
         logging.info("Init mod")
         if Mod.__single:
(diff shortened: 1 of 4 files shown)


rreichenbach (commented):
This looks good and seems to work quite nicely

PR #25, opened 2021-11-06
System Update menu item

micahvdm:
This is just the basic logic of showing the "System Update" option if hotspot is disabled. It then runs a simple bash script to update and the git pull. I have not added the pipreqs section as I have not looked into it yet, but should be fine to add at the end of the script.


micahvdm:
The logic I have used for "System Update" to only show when hotspot is disabled will not be affected by PR #23 as that is a line that is still being used to show the enable or disable hotspot option.

micahvdm:
Putting this on the back burner for now

PR #26, opened 2021-11-06
System Update fixed

micahvdm:
There was a versioning bug with adafruit_blinka and rtmidi got added instead of python-rtmidi.
This is all fixed now and working perfect. The requirements file can be updated with additional and the update system with git pull and then run the pip3 install -r requirements.txt. This is the safest way I have found to keep things in check when updating. If really necessary, we can add specific versions to each python package, but from what I've found, Pillow is the only one that can't go above 8.4.0 as some stuff breaks due to Mod specific issues.

### DIFF
--- a/pistomp/pistomp.py
+++ b/pistomp/pistomp.py
@@ -147,6 +147,7 @@ class Pistomp(hardware.Hardware):
                 timeout = 1000  # 10 seconds
                 initial_value = GPIO.input(f[2])
                 while self.test_pass is False and timeout > 0:
+                    fs.poll()
                     new_value = GPIO.input(f[2])  # Verify that LED pin toggles
                     if new_value is not initial_value:
                         break
@@ -240,5 +241,5 @@ class Pistomp(hardware.Hardware):
             GPIO.cleanup()
             sys.exit()
 
-    def test_passed(self, data):
+    def test_passed(self, data = None):
         self.test_pass = True


micahvdm:
This also does not get affected by #23 at all.

rreichenbach (commented):
This should probably be fine.  I want to test it well though on both Core and v1.  One thing I do know is, update.sh needs to have executable permissions.

rreichenbach:
Starting from a fresh build, then running update (after the permissions were fixed on update.sh), causes mod-ala-pi-stomp to fail:

Nov 08 07:33:23 patchbox python3[987]: AttributeError: module 'board' has no attribute 'CE0'

The update operation updated the board module to 1.0.  This PR will either need to peg the board version to the fresh-build version, or else change the pi-stomp code to work with the new board module version.

ozbenh:
I would say the latter (or rather work with both).

micahvdm:
This can be fixed by removing the version for Blinka. Blinka is actually what breaks it and not board

modalapi/mod.py
```diff
@@ -774,7 +776,14 @@ def system_disable_hotspot(self):
 
     def system_enable_hotspot(self):
         self.system_toggle_hotspot("Enabling, please wait...", "/usr/bin/patchbox wifi hotspot up")
-
+        
+    def system_updater(self):
+        self.update_system("Updating system, please wait...", "/home/patch/pi-stomp/modalapi/update.sh")
```
ozbenh, line 781: We should get rid of all those hard wired path... maybe dig them out of config, so we can one day start installing properly

modalapi/mod.py
```diff
@@ -783,6 +783,9 @@ def system_updater(self):
     def update_system(self, msg, cmd):
         self.lcd.draw_info_message(msg)
         subprocess.check_output(cmd, shell=True)
+        time.sleep(2)
```
ozbenh, line 786: Why the sleep ?

ozbenh:
Could you add something so that the update.sh scripts run a post-update.sh script from the pulled repository ? (if it exsits)  That way if we have extra requirements that aren't just pip (such as pulling new distro packages etc...) we can do it from there.

For example I want to add stuff there that checks if pisound-btn is installed and patchbox-wifi is available, and if true, remove the former and install the latter.

Also I didn't see in your code, do you restart the service after an update ?

rreichenbach:
So no, the board error is definitely because of it's reference in requirements.txt (not Adafruit_Blinka).   I tested 3 times from a fresh patchbox install with your cloned master branch.  Everything is fine until update.sh is run.  After that, restarting mod-ala-pi-stomp fails with the board error.  The third time I tested, I ran each pip3 install separately.  It only fails after the board install.  The board module is unique in that it's a preinstalled module.

As I said before, this functionality has no urgent need, so let's keep it shelved for now.  Eventually, it will need:
- board removed from requirements.txt and/or code change to work with version 1.0
- Execute permissions given to the update.sh script
- Hard coded path to update.sh replaced with self.homedir
- After update_system() is done updating, mod-ala-pi-stomp should be restarted.  Personally, I think the user should be prompted with a confirm before commencing with the update, but granted, there's not currently LCD/Nav support (like a Dialog or Confirm LCD object) to make that easy.

ozbenh:
>> but granted, there's not currently LCD/Nav support (like a Dialog or Confirm LCD object) to make that easy.

I think we should rework the screen management to more easily allow for such things, and alternate screen/menus etc... but it's not a trivial task. Do you already have plans/ideas in that area ? The main difficulty for people like myself is that I don't have a way to test with the other board model(s) with different LCD(s).

My thinking is to do a very simple "UI framework". I think we can, at least initially, stick to "everything is full screen", ie, each menu/dialog/... is a full page with the exception of the top bar with the network status.

There's a base class for such things and the LCD module keeps track of the "current" one. The page has a "refresh" method to redraw itself and an "input_event" method to handle rotary/button.

The next step is ... do we have a hierarchy of "Widget" objects ? (A very simple one). In which case we could easily have page pointing to the top level, and draw them top-to-bottom on refresh. All simple rectangles.

Then we need to figure out how we do the specialization of those for specific LCD types/models. Food for thought... do you want to discuss that on the forums ? another issue ? Leave it alone ? :-) Or a call/meeting to discuss live if you prefer.

I've written such things in past lives, I could try to sketch something out outside of pi-stomp first (I need to learn a bit more about how that LCD works), and maybe see if I can make it fast ...

ozbenh:
Regarding the update process, my thinking (feel free to disagree, it's your project :-) is that we should treat pi-stomp as a "product" and thus pin ourselves to known tested versions of dependencies.

That means that an "update script" would first grab the new version of pi-stomp hoping from release tag to release tag rather than whatever's top of tree on github (that way you have a bit of staging time before you tag new changes, more adventurous users can play with HEAD and more normal users stick to release tags).

The script would dig out some king of requirement file as @micahvdm did, but that file would contain *exact* component versions that are known to have been tested and work. IE. it's up to you/us to update that file when you/we specifically want to update a component, along with a pi-stomp release.

We could in fact do the same with dpkg packages. Ie, while we would do apt-get update to get up to date repo state, and we might get updates implicitely if we install new packages that have updated dependencies, the normal process is to specify versions. This is a bit trickier as it makes it harder to manage security updates, so it really depends how "product-ish" you want to be here.

Finally, the update script should run a generic "post update" script from the downloaded/updated pi-stomp that can be used to apply any new fixup that we deem necessary.

I would leave the responsibility of that post-update script to restart the service in case there are some specifics tricks to that new version involved in that restart process.

The update script could have command line options to bypass those version checks of course for advanced users.

rreichenbach:
Thanks for your input, Ben.  Yes, like you, I've been thinking more strict versioning is probably becoming a necessity.  The dependencies of the pi-Stomp specific code is rather small, but I have worried about the implications of not having tagged releases and everyone potentially running slightly different software.  Also that a package will update and break new installs.  My practice so far to catch such issues, is simply to test fresh builds almost weekly.

I like what you've laid out.  I do think that would be effort well spent.  I may have time in a month or two, but currently I am consumed with other things.  If either of you have the time/will to work on it, feel free, otherwise I'll prioritize it for end of the year/beginning of next.

Regarding the LCD... I'm not super proud of the LCD code.  It was a code-as-you-go effort and certainly a more scalable "framework"/OO design would be nice for new features.  Would the typical user see much benefit?  Probably not much.  Full screen refreshes will not be performant in my experience.  I settled on horizontal "zone" refreshes to avoid re-drawing images unrelated to the current event.  If every display object was a proper object having its own tiled image and refresh(), I do think that might work well.  A state model similar to React would be nifty.  

Despite it's flaws, I don't have a huge desire to rewrite a fairly complex thing that works fairly well for the most part.  If there was a significant back log of bugs or requested LCD features, then maybe, but I've already put way more time into it than I'd like to admit so, left to me, it'll be in maintenance mode until I get some free time.  I don't wish to discourage others from working on it though.  If you do wanna go down that road Ben, we should probably discuss it via Zoom.

ozbenh:
For updates I think the starting point is to start tagging releases (and decide what form that takes), then we can add the rest of the bits progressively. Maybe @micahvdm is interested, he has a "good start" even if it's not quite what it needs to be yet.

For the LCD, well, it's fun stuff so maybe .. :-) I don't know much about React ... pointers ? Otherwise I might just spend a bit of time toying around outside of pi-stomp environment see if I can come up with something that makes sense, then we can see if it's worth retrofitting. But by all mean let's chat in Zoom if you have time.

The thoughts I was bouncing in my head was basically divide the screen into "Page" stacks (one for the status bar, one for the main portion, at least initially). A page stack is just that .. a stack. ie, main page stack initially contains only the "main" page with all the usual controls, if you open the system settings, you "stack" up the system settings page which becomes the active page for that stack, when it closes, you go back to the previous one (pop the stack).

The widget selection iterates "selectable" widgets when turning the encoder in the current top of stack(s) and send button events to it (or to the widget itself, we'll see how I deal with that).

To deal with LCDs, I was thinking the widgets are created using a name string to a "factory" provided by the LCD which returns the right subclass for that LCD (or method tuple to stash in a generic subclass, we shall see).

That way we can much much more easily start adding "configuration" features that don't require a laptop and a web browser, which I can see being useful on site (oops I forgot to attach that control to that pot !). And I've been musing having a wifi config UI since getting to AP mode and using ssh is ... a PITA for a non-geek user.

That "page stacks" idea makes that somewhat easier as for example the main "mod" handler doesn't need to worry about what the "wifi config" page is doing, it's a separate object, it just happens to be the current top of stack, so it will get the encoder clicks.

As for refreshes, a simple hierarchy of widgets with basic rectangle "zones" makes it rather easy to limit the LCD refresh to object that changed.

Anyways, those are just idle thoughts as I was sitting in the tram for too long today :-)

PR #28, opened 2021-11-22
[RFC] pistomp v1: Fix footswitch hardware test

ozbenh:
NOTE: I don't have a v1 to test, this is based on copying some of that test code into pistompcore.py
and hard wiring the setup. Hopefully that's all that is broken.

test_passed can be called without argument and we need to poll
the footswitch to get events

Signed-off-by: Benjamin Herrenschmidt <benh@kernel.crashing.org>


rreichenbach:
Tested on v1 hardware.  Yup, that's it.  Thanks!

PR #43, opened 2022-06-29
Low Latency Kernel

micahvdm:


rreichenbach:
The good news is that, yes, this seems to fix the USB issue.  The bad is that CPU usage is back to what it was before the real-time kernel: 50 to 70% for most of the default pedalboards, where with the RTK those same pedalboards used about half (25 to 35%).  Personally, I'd even give up USB for the better performance and almost zero XRUNs.  Might be better to fix the USB issue with the RTK if possible.

micahvdm:
I managed to build 5.15.40-rt43 kernel and it works now when adding dtoverlay=dwc2,dr_mode=host to config.txt.
It works with the previous kernel too, but wasn't as stable. New kernel is much better, stability wise.

rreichenbach:
Seems good.  CPU usage is back to the 30%'s and I'm able to hot plug a USB device with apparently no ill effects.

PR #48, opened 2022-08-09
-S forces 16bit mode

micahvdm:
This significantly reduces the audio quality, leads to ugly clipping.


rreichenbach (approved):
Looks good and verified performance with AudioInjector card

PR #51, opened 2022-09-11
64 bit rtk

rreichenbach:
This commit switches from a 32-bit PatchboxOS base to use 64-bit Raspberry Pi OS lite with realtime kernel.  Patchbox/MODEP included MOD v1.8.  We now can use MOD v1.12.


micahvdm (approved):
It all looks good to me!

PR #58, opened 2023-03-26
add symlink, update kernel and fix IQAudio card

micahvdm:
Symlink is for pedalboard presets issue
Updated kernel using a newer GCC
The IQAudio fix unfortunately only works persistently if the user uses the shutdown or restart options from the system menu.
I have not been able to get it to work any other way yet, as the state for alsactl seems to load very early in the boot process.
This is why we need to disable the aux jack pin at shutdown and store it to state. By the time the rc.local file loads, the PLL has settled and the correct sample rate is configured. I have reported this to RPi on the kernel repo.


rreichenbach:
A change is needed for setup/pi-stomp-tweaks/modify_version.sh which gets called at the beginning of install.  It needs the new paths and a mkdir:

-config_file="$HOME/pi-stomp/pistomp/default_config.yml"
-pistomp_orig_config_file="$HOME/pi-stomp/pistomp/default_config_pistomp.yml"
-pistomp_core_config_file="$HOME/pi-stomp/pistomp/default_config_pistompcore.yml"
+config_dir="$HOME/data/configs"
+config_file="$config_dir/default_config.yml"
+pistomp_orig_config_file="$HOME/pi-stomp/setup/config_templates/default_config_pistomp.yml"
+pistomp_core_config_file="$HOME/pi-stomp/setup/config_templates/default_config_pistompcore.yml"
 
+mkdir -p $config_dir

rreichenbach (approved):
looks good

PR #62, opened 2023-05-10
Update to Mod-UI 1.13 with Patchstorage

micahvdm:
This moves us off of Mod official repo to mine with our patches applied and patchstorage implemented. Jack2 is also moved to my repo instead of Mod repo as there seems to be a midi clock sync bug with Mod one. I have a patch added for Jack2 and the timing is fixed. Tested and working.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -21,21 +21,23 @@
 #
 usage()
 {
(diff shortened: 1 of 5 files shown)


rreichenbach (approved):
All tested, seems to work and looks good.

PR #63, opened 2023-05-17
Fix MIDI issues by adding mod-midi-merger back

micahvdm:
Tested and working


rreichenbach (approved):
changes tested and look good.

PR #68, opened 2023-06-27
Override menu width

micahvdm:
This prevents title text overflows


micahvdm:
I've commented out the width adjustment for now and set w to a constant. This looks good across the board so far.

PR #71, opened 2023-07-08
Add back audio card fixes for shutdown and restart

micahvdm:
These are in mod.py but not in modhandler.py

### DIFF
--- a/modalapi/modhandler.py
+++ b/modalapi/modhandler.py
@@ -357,15 +357,13 @@ class Modhandler(Handler):
     def system_menu_shutdown(self, arg):
         self.lcd.splash_show(False)
         logging.info("System Shutdown")
-        os.system('sudo systemctl --no-wall poweroff')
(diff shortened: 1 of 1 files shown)


rreichenbach (changes requested):
The amixer command is specific to IQAudio, it will fail for other cards.  A main reason that the Audiocard classes exist for each card type, is to abstract the details for such operations.  Polymorphism.

instead of a system call, self.audiocard.set_parameter() should be called with a generic parameter name that's made specific for each card by that card's subclass.   This is currently done for CAPTURE_VOLUME and MASTER.  There doesn't seem to be a mute for AudioInjector, so how about just setting MASTER to zero?  

Similarly, store and restore should be done by calling self.audiocard.store()

Those methods also include try/except so we don't have to worry about a failed system call causing a crash.

If a similar operation other than set, get, store and restore is needed, that could be added to the AudioCard base class.

micahvdm:
This will no longer be needed with the kernel level fixes

PR #75, opened 2023-07-25
update to new 5.15 kernel with permanent samplerate fixes

micahvdm:
also removing old hacks for iqaudio samplerate fix


rreichenbach:
Looks good and works good

rreichenbach (approved):

PR #79, opened 2023-09-08
Check wifi service

micahvdm:
This service runs at boot after networking service has completed. It checks iwconfig for an ESSID and ignores if found. If not found, then it will start hotspot mode


micahvdm:
Closing until it's working correctly

PR #83, opened 2023-09-18
Wifi power save fix

micahvdm:
Wifi power save gets enabled before network gets a chance to stabilize, causing the wifi_check script to default to hotspot mode almost every time. This fixes it as the network interfaces conf runs before rc.local allowing the network to connect cleanly.


rreichenbach (approved):
Seems fine

PR #94, opened 2023-11-06
Updates for toolbar icon statuses

micahvdm:
This PR adds in hotspot toggle to the wifi dialog. It also handles the eq status and wifi status for the icons.
I also added in the footswitch definitions to the tre config file for easier reference.


pistomp/lcd320x240.py
```diff
@@ -179,28 +179,43 @@ def draw_tools(self, wifi_type=None, eq_type=None, bypass_type=None, system_type
...
-        b = TextWidget(box=Box.xywh(0, 0, 0, 0), text='mySSID', prompt='SSID :', parent=d, outline=1, sel_width=3,
-                       outline_radius=5,
-                       action=lambda x, y: self.pstack.pop_panel(d), align=WidgetAlign.NONE, name='cancel_btn',
-                       edit_message='WiFi SSID')
-        d.add_sel_widget(b)
-        b = TextWidget(box=Box.xywh(0, 30, 0, 0), text='password123', prompt='Password :', parent=d, outline=1,
-                       sel_width=3, outline_radius=5,
-                       action=lambda x, y: self.pstack.pop_panel(d), align=WidgetAlign.NONE, name='cancel_btn',
-                       edit_message='Password')
+        self.w_wifi_ssid = TextWidget(box=Box.xywh(0, 0, 0, 0), text='mySSID', prompt='SSID :', parent=d, outline=1, sel_wi>
```
rreichenbach: Typo here, looks like like got chopped

modalapi/wifi.py
```diff
@@ -117,12 +117,46 @@ def enable_hotspot(self):
...
+            'country=US',
+            '\n'
+            'network={',
+            '\tssid="{}"'.format(ssid),
+            '\tpsk="{}"'.format(password),
+            '}'
+        ]
+        config = '\n'.join(config_lines)
+
+        os.popen("sudo chmod a+w /etc/wpa_supplicant/wpa_supplicant.conf")
```
rreichenbach: This file path is referenced in 3 places, too easy to miss one or typo, if it ever changes.  Please declare it in one place and then reference the variable each place.

modalapi/wifi.py
```diff
@@ -117,12 +117,46 @@ def enable_hotspot(self):
...
+        logging.info("Wifi config added. Refreshing config")
+
+        os.popen("sudo wpa_cli -i wlan0 reconfigure")
+
+    def get_wifi_credentials(self, file_path):
+        file_path = '/etc/wpa_supplicant/wpa_supplicant.conf'
+        with open(file_path, 'r') as file:
+            for line in file:
+                if 'ssid=' in line:
+                    current_ssid = line.split('"')[1]
```
rreichenbach: I feel like this line could fail if the file were hand edited.  For example, I believe single ticks around strings is still valid.  At a minimum, it should not fail with a "list index out of range" if the string can't be parsed.

rreichenbach (commented):
A few fairly minor comments...

micahvdm:
I have fixed the above stuff, please check.
I also removed the get WiFi credentials part as it could indeed cause issues…

micahvdm:
Will reopen when all is fixed...

PR #99, opened 2024-06-22
Initial updates for Wifi in bookworm

micahvdm:
This updates the hotspot code to use nmcli instead of wpa_cli and dhcpd. It also adds the base connections routines for the wifi connection via ui.


rreichenbach (approved):
Tested these changes.  Hotspot does at least start, but the connection is almost unusable.  I'm guessing some re-configuration might be required.

PR #102, opened 2025-01-21
Pi5 Eeprom update script

micahvdm:
This is to allow the pi5 to deep sleep and turn off the gpio and 3.3v rail, which turns off the display when shutting down with either the power button or the shutdown menu item. This might also work on pi4, but I don't have a pi4 to test on currently. It's only really helpful on pi4 to see that the device is completely shutdown.


rreichenbach:
In general, most "rm" commands should include a "-f" so the whole script doesn't fail if the file is missing.  Probably not a problem here.

PR #109, opened 2025-12-22
External MIDI devices and MIDI passthrough

sastraxi:
> Please note that I'm sharing this not to get immediately merged, but to start a discussion.

First off: what an incredible pedal. I've been having so much fun. I'm a bassist, and my plan for this pedal is mainly to use it as the "last mile" of signal changing. I've already got a really competent board to make my bass sound, well, like a bass should. And I use the Source Audio C4 pedal a lot for its excellent tracking and customizability. I followed your instructions to create an expression pedal input with my M-Audio EX-P, and it works flawlessly despite my lack of soldering skills.

Here's my signal flow now that I've built the piStomp (note the Sansamp BDDI is simply being used as an active Y-splitter):

Signal flow

This is where I met up with a bit of resistance. The features I didn't have:

1. I wanted each pedalboard to be assigned to a specific C4 preset (you can load 128 of them and then switch 'em via MIDI CC)
2. While I could assign the expression pedal through MOD pedalboard to any of the controls there, I couldn't actually forward those MIDI events to the C4
3. Similarly, I wanted to be able to control aspects of the C4's preset (e.g. turning on/off an LFO, changing wet/dry) using the footswitches, tweak encoders, and tweak encoder buttons.

You can read the `CLAUDE.md` which outlines the changes I've made. These are:

1. Add an `external_midi.yml` configuration which lets you define expected MIDI devices (i.e. via USB MIDI class) and automatically send MIDI messages to them when a certain pedalboard is loaded.
2. Create a virtual `piStomp-MIDI` port via `amidithru`, recognized by MOD pedalboard to send MIDI events to, rather than using the ALSA default "MIDI Through". This way we can forward / filter / modify the ones I want via the MIDI routing in the pedalboard, and the parameters still work for MIDI Learn.
3. For ADC inputs mapped to MIDI (like the expression pedal), I added sync functionality that sends a single MIDI event when the pedalboard is loaded, so that the current value of the expression pedal is used immediately.
4. Add a `shortpress` config in the YAML and the same callback approach that `longpress` has so I can generate MIDI CCs via the short press of the Tweak1/Tweak2 encoders rather than their current behaviour of doing the same thing as clicking the Navigation encoder.

These four features gave me everything I needed, combined with the existing ability to use the MIDI Clock Generator plugin to ensure BPM flows to the C4 via the MOD pedalboard graph.

Let me know if you have any questions. Obviously it's a big ol' monolithic hunk of code that probably doesn't follow your intended style and, admittedly, was written mostly using Claude Code (though I tried to ape the style as much as possible while adding some type safety). I'm curious if you find any value in these features, or if you'd suggest approaching them differently. As it is right now, I'm having a lot of fun and everything's working quite well.


maarthome (commented):
Hi,

Did you succeed in compiling pi-stomp in Debian 13?
If so.... how?

Cheers,
Maarten

sastraxi:
@maarthome I haven't tried that yet -- I simply used the prebuilt xz image (https://github.com/TreeFallSound/pi-stomp/releases/tag/v3.0.3) and copied my changes over via ssh.

rreichenbach (commented):
This is very cool.  Lots of good stuff in there.  You've hit some things that have been on my ToDo list for a while.  My concern for a change like this is regression, example: interference with the event loop since those kind of problems can manifest intermittently.  I'll try to review and test this fully by mid January.  Will need to test on all 3 hardware generations.

In the future, it would be easier to review and test if each PR includes a single feature/function.   In this case, I get that they are all MIDI related and work in tandem to give you the functionality you wanted.  Might need to review one commit at a time to understand the intent of each.

Claude did a pretty good job of summarizing the codebase.  Similar to what I created here: https://treefallsound.com/wiki/doku.php?id=developers_guide    Would at least like to rename that file to something like CODE_ANALYSIS.md

The toml/uv.lock will likely conflict with the dependencies defined in pi-gen-pistomp.  Who is Trevor Willig?  Zat you?

sastraxi:
@rreichenbach haha this is why I mentioned a lot of it was vibe coded (and I didn't even mean to commit the pyproject / uv.lock) -- Claude happily made up that name 😂 If you haven't used uv I'd definitely suggest looking into it for the python dependency management, but it should be a separate PR as well if you decide to go down that route. Can you point me towards pi-gen-pistomp?

Regarding `CLAUDE.md` to `CODE_ANALYSIS.md`, I agree. No harm in just importing that file when we're using coding assistants.

I definitely don't want to merge this as is -- I'll split it up into logical increments and open individual PRs that build on each other. Regarding the event loop, I'll ask for these smaller reviews when I split them up, should be easier to tell what needs to be done.

The biggest change here (IMO) is the move from using `Midi Through` to creating our own `piStomp-MIDI` virtual port. I worry what that might break, but my reading of the mod-host codebase says it should be fine for MIDI learn (it picks up MIDI port creation events, thankfully).

sastraxi:
Closing in favour of #110 (and further stacked PRs).

maarthome:
I really would appreciate how to use your additions... I am no programmer... but I see the benefit of sending  configuration parameters after pressing a foot switch... to another device...
So a short DIY is really  helpful to me...

Thanks for putting so much energy in the pistomp project... I love this application...

Cheers,
Maarten

sastraxi:
@maarthome sure thing! If you download this branch (or better, https://github.com/sastraxi/pi-stomp/pull/5) you can ssh into your piStomp and copy the changed files over. Then, it's a matter of customizing your config files and restarting the device. You can take a look at the example YAML here for how to change things around.

maarthome:
Hi Cameron,

I do basically understand what al the branches are. But I do not understand how they are related.
Or... which version is the lastest working version?
Until now I just pulled branches with the the most commits ahead of the original 'TreeFallSound/pi-stomp:pistomp-v3' version but they do not always work....
So... how do I get the latest and greatest vesion you made?

Thanks for all your great work!! Very nicely detailed!!
I hope Randall will incoporate your additions and modifications. This could be maintained in a separate development branch at The ThreeFallSound repository.

Cheers and again thanks for your time and energy!

Maarten

sastraxi:
The PRs have been split up for easier review and testing. I’m maintaining the originals on my fork; if you’re interested in the latest code I think the branch name is feat/collage-mode

sastraxi:
Reopened by accident

PR #111, opened 2026-01-07
Prevent crashes when update_bypass is called too early

sastraxi:
I got a crash on changing pedalboards that said `self.w_power` was `None` -- caused by `update_bypass` being called before the power icon had been (re-)created.


rreichenbach (approved):
Nice catch.  Looks good.

PR #112, opened 2026-01-07
Only pop the main panel if it was actually pushed

sastraxi:
Fixes a bug I was seeing when rapidly switching between tweak encoders and volume changing.


rreichenbach (approved):
Looks good.

PR #113, opened 2026-01-07
Control other devices/pedals via MIDI

sastraxi:
Any hardware control can now send its MIDI to an external device port (Source Audio C4, HX Stomp, etc.) instead of the virtual MIDI Through port. Also introduces a pedalboard-load message sending mechanism, allowing pedalboards to ensure external devices are in known state.

Includes https://github.com/TreeFallSound/pi-stomp/pull/169

### What's new

1. The main feaure is per-control routing. Add `midi_port: <MIDI device name>` to any footswitch, encoder, or analog/expression control, and its MIDI is sent there instead of to mod-ui, falling back if the device is unavailable.
2. An external MIDI manager class provides port enumeration/auto-detect, eager open at routing time, send-with-fallback wrapper, and a pedalboard-load message burst (send_messages_for_pedalboard) for syncing external gear on bundle change.
3. This branch introduces a controller manager class, shared by mod/modhandler, as well as extracting out Current into its own (shared) module.
4. Finally, external controllers show on the LCD, just like they would if they were assigned to a parameter in MOD. Externally-routed controls get a synthetic parameter to do so and work in all hardware versions.
5. Config files, parser, and hardware have been updated to pass-through configuration

### Testing

I've used this for a few months, both via init message and scrolling through presets on my C4 with Tweak1. This version has been simplified, decomposed, and tested more thoroughly than that one.

There are MacOS-only loopback tests that open an actual virtual MIDI port and assert bytes travel the full path for every routable control: footswitch press, tweak-encoder rotation, expression movement, plus the pedalboard-load messages and backoff.
 
### Bugs discovered (mod-host)

Connecting/disconnecting an rtmidi output can trigger a crash in mod-host while it attempts to clean up when changing pedalboards. It's relatively rare and I only managed to trigger it by having a pedalboard with 54 (!) dead MIDI port references in the TTL. This will be worked on separately (see https://github.com/mod-audio/mod-host/pull/96)

> When `ExternalMidiManager._init_port()` opens a direct rtmidi port to the C4, it creates
> a new ALSA sequencer client (`RtMidiOut Client`, client 132). JACK sees this as a new
> `JackPortIsTerminal|JackPortIsPhysical` MIDI port and fires mod-host's async
> `PortRegistration` callback, which enqueues a `POSTPONED_JACK_MIDI_CONNECT` event.
>
> If the user switches pedalboards before that event is drained, mod-host's main thread is
> simultaneously running `effects_remove()` → `lilv_instance_free()` → `dlclose()`. The
> async connect event and the teardown race on the ALSA/JACK heap, causing corruption.
> Subsequent `dlclose()` calls then SIGSEGV on the corrupted state.
>
> Sonnet 4.6


modalapi/modhandler.py
```diff
@@ -91,24 +94,37 @@ def __init__(self, audiocard, homedir):
...
+
         # Callback function map.  Key is the user specified name, value is function from this handler
         # Used for calling handler callbacks pointed to by names which may be user set in the config file
         self.callbacks = {"set_mod_tap_tempo": self.set_mod_tap_tempo,
                           "next_snapshot": self.preset_incr_and_change,
                           "previous_snapshot": self.preset_decr_and_change,
                           "toggle_bypass": self.system_toggle_bypass,
-                          "toggle_tap_tempo_enable": self.toggle_tap_tempo_enable
+                          "toggle_tap_tempo_enable": self.toggle_tap_tempo_enable,
+                          "universal_encoder_sw": self.universal_encoder_sw
```
sastraxi: Need to remove this

typings/rtmidi/_rtmidi.pyi
```diff
@@ -0,0 +1,31 @@
+from typing import Callable, Sequence
```
sastraxi, line 1: All these typings will be removedf

modalapi/mod.py
```diff
@@ -156,6 +158,13 @@ def __init__(self, audiocard, homedir):
         self.ws_bridge.start()
         logging.info("WebSocket bridge started")
 
+        # External MIDI device synchronization
+        self.external_midi = None
+        try:
+            self.external_midi = ExternalMidi.ExternalMidiManager()
+        except Exception as e:
+            logging.warning(f"Failed to initialize external MIDI manager: {e}")
+
```
sastraxi, line 163: No reason to try...except. Ctor will not throw

modalapi/mod.py
```diff
@@ -608,6 +611,17 @@ def set_current_pedalboard(self, pedalboard):
         self.load_current_presets()
         self.update_lcd()
 
+        # Send external MIDI messages for this pedalboard
+        # Config was already updated by hardware.reinit(cfg) above
+        if self.external_midi is not None:
+            try:
+                self.external_midi.send_messages_for_pedalboard()
+            except Exception as e:
+                logging.warning(f"Failed to send external MIDI messages: {e}")
```
sastraxi: Don't need if guard

pistomp/analogmidicontrol.py
```diff
@@ -64,55 +52,45 @@ def set_midi_channel(self, midi_channel):
...
+        if value <= self.tolerance:
+            return 0
+        if value >= 1023 - self.tolerance:
+            return 1023
+        return value
 
-        # read the analog pin
-        value = self._clamp_endpoints(self.readChannel())
+    def _send_value(self, value):
+        """Send ADC value as MIDI CC and invoke callback if set."""
```
sastraxi, line 64: I think this is actually worse than the initialize() method value. Likely it's reverting to an earlier version of the code.

PR #114, opened 2026-01-07
Sync current state of analog controls on pedalboard load

sastraxi:
When using e.g. expression pedals with MOD pedalboard, this PR adds an initial "sync" where we read the value of the GPIO pin and send a MIDI message so that we don't have to e.g. wiggle the pedal to get it to reflect the current state of the analog input.

### DIFF
--- a/pistomp/hardware.py
+++ b/pistomp/hardware.py
@@ -88,6 +88,18 @@ class Hardware:
         if s:
             s.check_longpress_events()
 
+    def sync_analog_controls(self):
+        """
+        Send current values of all analog controls via MIDI.
+        Used for syncing external devices during pedalboard load.
+        """
+        for control in self.analog_controls:
+            if hasattr(control, 'send_current_value'):
+                try:
+                    control.send_current_value()
+                except Exception as e:
+                    logging.warning(f"Failed to sync analog control {control.midi_CC}: {e}")
+
     def poll_indicators(self):
         for i in self.indicators:
             i.refresh()
--- a/pistomp/hardware.py
+++ b/pistomp/hardware.py
@@ -20,6 +20,7 @@ import sys
 
 import common.token as Token
 import common.util as Util
+from pistomp.analogcontrol import AnalogControl
 import pistomp.analogmidicontrol as AnalogMidiControl
 import pistomp.footswitch as Footswitch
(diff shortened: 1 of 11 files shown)


rreichenbach (approved):
This is cool. I can't merge it though since this fork/branch needs a rebase.

renemadeira:
This is very cool — thanks for the implementation.

Since my unit has a lot of analog controls, I was wondering if this behavior could be made optional. For example, a per-control toggle in the analog control definition (something like autosync=1).

For some controls it’s great to sync immediately on preset/pedalboard change, but for others I’d prefer to keep the value stored in the snapshot until the control is moved.

From a usability standpoint, having this as a per-control option would cover both use cases cleanly.

Thanks again for the great contributions.

sastraxi:
@renemadeira added! The default is `autosync=False`, but documented as the recommended value for the expression pedal in the default configs

@rreichenbach I re-did this PR based on the above feedback and also realizing that we can better encapsulate the controls by making them responsible for their own initialization. Let me know if you'd like any refactors.

pistomp/hardware.py
```diff
@@ -104,13 +105,15 @@ def reinit(self, cfg):
         # reinit hardware as specified by the new cfg context (after pedalboard change, etc.)
         self.cfg = self.default_cfg.copy()
 
-        self.__init_midi_default()
```
sastraxi, line 107: This just does the same as the "pedalboard-specific config" below.

pistomp/hardware.py
```diff
@@ -104,13 +105,15 @@ def reinit(self, cfg):
...
-
         # Global footswitch init (callbacks and groups)
         Footswitch.Footswitch.init(self.handler.callbacks)
 
-        # Footswitch configuration
-        self.__init_footswitches(self.cfg)
```
sastraxi, line 113: Same for this; already done below.

pistomp/hardware.py
rreichenbach, line 107: No, this is required for popper initialization of default config

pistomp/hardware.py
rreichenbach: No, this is required for popper initialization of default config

PR #129, opened 2026-04-18
Ensure analog inputs can reach their ends

sastraxi:
The tolerance deadband (used to suppress jitter near the center; hysteresis) also prevented the ADC from ever reporting exactly 0 or 1023 at the extremes. Expression pedals couldn't reach their full range — they'd always stop a few counts short of the endpoints.

This was especially observable when e.g. an expression pedal was moving quickly.


rreichenbach (approved):
Looks good.  Tested and does seem to work better for my expression pedal.

PR #130, opened 2026-04-18
Pillow dependency and 10+ compatbility

sastraxi:
* added pillow to `pyproject.toml` (>= 9.4, which is what is currently used)
* moved from `getsize` to `getbbox` (former was deprecated)
* fix for degenerate rectangles causing problems in Pillow 10+
* micro-optimization: do not refresh text if we set it to the same string

#### Branch stack
* https://github.com/TreeFallSound/pi-stomp/pull/130
  * https://github.com/sastraxi/pi-stomp/pull/16


rreichenbach (approved):
Looks good, seems to work fine

PR #131, opened 2026-04-18
Drop black; add pyright; address some type issues

sastraxi:
We already have ruff, so black is superfluous.

Adds pyright for type checking, type stubs for untyped dependencies, and begins typing some of the core of pistomp-v3.


modalapi/modhandler.py
```diff
@@ -121,11 +122,35 @@ def __init__(self, pedalboard):
...
+
+    @property
+    def lcd(self):
+        assert self._lcd is not None, "LCD has not been initialized"
+        return self._lcd
+
+    @property
+    def hardware(self):
+        assert self._hardware is not None, "Hardware has not been initialized"
+        return self._hardware
```
sastraxi, line 155: These are guaranteed to be not None after hardware sets 'em, so this is a clean fix for these two. Ideally we'd resolve the circular dependency between Hardware and the handler classes, then we don't need this hack.

modalapi/modhandler.py
```diff
@@ -202,16 +227,15 @@ def poll_system_info(self):
...
     def poll_lcd_updates(self):
-        if self.lcd:
-            self.lcd.poll_updates()
+        self.lcd.poll_updates()
 
     def universal_encoder_select(self, direction):
-        if self.lcd is not None:
-            self.lcd.enc_step(direction)
+        if self._lcd is not None:
+            self._lcd.enc_step(direction)
```
sastraxi, line 237: For those methods that are / can be called directly from Hardware, we use `_lcd` directly so we don't depend on startup ordering.

modalapi/modhandler.py
```diff
@@ -671,22 +682,23 @@ def user_restore_data(self, arg):
             logging.info("No USB device found")
             self.lcd.draw_message_dialog("No USB device found")
 
-    def system_menu_save_current_pb(self, arg):
+    def system_menu_save_current_pb(self, _arg: None):
+        if self.current is None:
+            logging.error("No current pedalboard set, cannot save")
+            self.lcd.draw_message_dialog("No current pedalboard set, cannot save")
```
sastraxi, line 689: Maybe controversial

pistomp/config.py
```diff
@@ -130,6 +129,9 @@
               },
               "type": {
                 "enum": ["KNOB", "EXPRESSION"]
+              },
+              "autosync": {
+                "type": "boolean"
```
sastraxi, line 134: Missed this in a previous PR

pistomp/hardware.py
```diff
@@ -268,9 +275,11 @@ def create_analog_controls(self, cfg):
             logging.debug("Created AnalogMidiControl Input: %d, Midi Chan: %d, CC: %d" %
                           (adc_input, midi_channel, midi_cc))
 
-    def add_encoder(self, id, type, callback, longpress_callback, midi_channel, midi_cc):
+    @abstractmethod
+    def add_encoder(self, id, type, callback, longpress_callback, midi_channel, midi_cc) -> Encoder.Encoder | EncoderMidiControl.EncoderMidiControl:
         # This should be implemented by hardware subclasses that support tweak encoders (Tre at least)
-        pass
+        ...
```
sastraxi, line 281: This is the right way of doing abstract methods according to the python docs

pistomp/hardware.py
```diff
@@ -15,20 +15,25 @@
...
 import pistomp.taptempo as taptempo
 
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 
+Controller = Union[AnalogMidiControl.AnalogMidiControl, EncoderMidiControl.EncoderMidiControl, Footswitch.Footswitch]
 
-class Hardware:
+
+class Hardware(ABC):
```
sastraxi, line 36: Needed for pyright to not complain about the return type of the local method

modalapi/modhandler.py
```diff
@@ -19,40 +19,41 @@
...
 from pistomp.footswitch import Footswitch
-from pistomp.handler import Handler
 from pathlib import Path
 
 
 class Modhandler(Handler):
     __single = None
 
     def __init__(self, audiocard, homedir):
-        self.wifi_manager = None
```
sastraxi, line 43: Defined authoritatively later in this constructor

modalapi/modhandler.py
```diff
@@ -202,16 +229,15 @@ def poll_system_info(self):
             self.temperature = "unknown"
 
     def poll_lcd_updates(self):
-        if self.lcd:
```
rreichenbach, line 205: I think we should still have the protective condition there:

if self._lcd is not None:
    self._lcd.poll_updates()

rreichenbach (changes requested):
This looks good.

I have a suggestion for one change.

sastraxi:
@rreichenbach addressed!

rreichenbach (approved):

PR #133, opened 2026-04-19
Move and extend `Parameter` to `common/`

sastraxi:
`Parameter` was living in `modalapi/` despite being a domain object needed well beyond the MOD API layer. This PR moves it to `common/` and extends it with the unit and taper metadata that later PRs require — so that downstream code can ask a parameter how to display and scale itself rather than repeating that logic at every call site.

- Relocate `modalapi/parameter.py` → `common/parameter.py`
- Add `unit_symbol` / `unit_label` parsing from TTL
- Add `get_taper()` and `format()` methods
- Update all imports across the codebase


rreichenbach (approved):
Pretty sure the modalapi/parameter.py shim could just be removed, right?

sastraxi:
@rreichenbach agh! I didn't mean to commit that. I'll get to that on Sunday morning.

PR #134, opened 2026-04-19
Replace wpa_cli with nmcli for Arch Linux compatibility

sastraxi:
Arch doesn't have `wpa_supplicant`, but we already depend on `nmcli`, so we can just replace the call with an equivalent one using that binary. Works on pi-gen-pistomp too.


rreichenbach (approved):
Awesome.  Seems to work.  Thanks.

PR #140, opened 2026-05-10
Rewrite wifi menu 

sastraxi:
## Wifi system overview

`modalapi/wifi.py` has been refactored into a modalapi.wifi package with five modules:

- types.py: shared data types
- nmcli.py: subprocess wrapper and nmcli output parsers
- ops.py: wifi operations that use nmcli under the hood
- commands.py: `Command[T]` and concrete command dataclasses (`ConnectSavedCmd`, `ConnectScannedCmd`, `ScanCmd`, etc.), plus a single-worker serialized `CommandQueue` executor
- manager.py: integrates with mod/modhandler, owns the polling thread, status cache, and queue instance

### Bugfixes

1. I suggest you apply https://github.com/TreeFallSound/pi-gen-pistomp/pull/12 to improve scan UX.
2. Includes https://github.com/TreeFallSound/pi-stomp/pull/146. This affects a lot of snapshots as it subtly changes how icons are rendered.

### Demo (out-of-date)

While not accurate to what the current wifi menu looks like, it's accurate in terms of functionality.

https://github.com/user-attachments/assets/a83bea19-2866-4454-9bc1-44be3211b591

### Author checklist

- [x] port over tests
- [x] end-to-end testing with emulator and debug calls
- [x] test on-device (pistomp-arch)


rreichenbach (commented):
This is super cool, but not ready.  Switching to Hotspot works, but I can't get it back to WiFi.  It sometimes wouldn't even show Nearby networks.  When it did, it made me re-enter a password for the original wifi network.  To avoid frustration, it should remember the password previously used so one can toggle back and forth between hotspot and wifi.

FWIW, I found that working/testing this kind of stuff gets really frustrating if you switch modes and rely on that connection for making your code changes.  It's good to have a wired connection that's always available.

sastraxi:
> This is super cool, but not ready. Switching to Hotspot works, but I can't get it back to WiFi. It sometimes wouldn't even show Nearby networks. When it did, it made me re-enter a password for the original wifi network. To avoid frustration, it should remember the password previously used so one can toggle back and forth between hotspot and wifi.
> 
> FWIW, I found that working/testing this kind of stuff gets really frustrating if you switch modes and rely on that connection for making your code changes. It's good to have a wired connection that's always available.

Randall, I apologise. I accidentally did not push up my latest round of changes 🙈 I'll see if I can repro your bug on the current tip.

sastraxi:
Fixed and added some tests for what you found.

rreichenbach (commented):
Better, but I'm still having problems using this.  When I first switched to Hotspot mode, the pi crashed.  I power-cycled, it did come up in hotspot mode, so from that point maybe ok.

But then I disabled Hotspot mode and it does remember my previous network and connect - that's good, but the dialog title shows "WiFi - Disconnected".  And if I choose that network, it will require I enter a password.  It would be nice if the currently connected was identified that way (eg. "YYZ (current)" or has an icon), and maybe even not allowing it to be clicked, so the user doesn't waste time entering a password not knowing they don't need to.

Another UX issue: it's not intuitively obvious to me that clicking "Hotspot Mode" will toggle it on or off.  I like the little dot indicator, but actually think it would be more clear if it showed "Switch to Hotspot Mode" when connected to WiFi, and "Disable Hotspot Mode" when in Hotspot mode.

Also when I click the wifi toolbar button, there's usually some significant lag time.  I believe it's finding the networks.  I wonder since that initial dialog doesn't show all the networks, if the network load could be lazy (after the dialog appears).

I know I'm asking for a lot of UX here and that's the UI lib isn't exactly robust.  But lack of connectivity can cause major frustration, so if we're going to improve this, it should work rather smoothly.

modalapi/mod.py
```diff
@@ -852,7 +854,7 @@ def system_info_populate_wifi(self):
             self.menu_items[key] = {Token.NAME: self.wifi_status[key], Token.ACTION: None}
             if self.wifi_status[key]:
                 hotspot_active = True
-        key = 'ip_address'
+        key = 'ip4_address'
```
sastraxi, line 857: Unrelated bugfix

pistomp/lcd320x240.py
```diff
@@ -306,16 +261,27 @@ def draw_preset_menu(self, event, widget):
...
         self.draw_selection_menu(items, "Snapshots", auto_dismiss=True, dismiss_option=True)
 
-    def draw_selection_menu(self, items, title="", auto_dismiss=False, dismiss_option=False):
-        # items is list of touples: (item_label, callback_method, callback_arg)
-        # The below assumes that the callback takes the menu item label as an argument
+    def draw_selection_menu(self, items, title="", auto_dismiss=False, dismiss_option=False,
+                            font=None, title_font=None, default_item=None):
+        # items is a list of tuples: (label, callback, arg) or (label, callback, arg, is_active)
+        # or (label, callback, arg, is_active, long_callback) where long_callback is called
+        # instead of callback on a long press.
```
sastraxi, line 268: This is getting pretty out-of-hand, will tackle in a future PR

rreichenbach (approved):
Seems to work great and usage feels intuitive.  I know that was a lot of work.  Thanks!

PR #141, opened 2026-05-10
Use websocket bridge for snapshot and pedalboard change events

sastraxi:
Rather than loading `last.json` and detecting modification time changes, we can load the current pedalboard using the REST API on startup (mod-ui starts before pistomp) and then track pedalboard changes using websocket messages.

We still use the polling loop for mod-ui changes, but instead of checking the file, we check for queued messages from mod-ui on the websocket since the last time we checked.

Bonus: we can actually read snapshot changes this way too, so this PR also adds that feature to sync the current snapshot if changed in MOD-UI.

#### Branch stack
* This branch (https://github.com/TreeFallSound/pi-stomp/pull/141)
  * https://github.com/sastraxi/pi-stomp/pull/15
  * https://github.com/sastraxi/pi-stomp/pull/25


rreichenbach (commented):
This is also very cool.  So stoked to see the Snapshot change from MOD-UI sync with the LCD.  It all seems to work, but I want to review the code better.  If for no other reason to understand how to use it.  I'm hoping this will eventually allow us to sync plugin bypass events too.

sastraxi:
> This is also very cool. So stoked to see the Snapshot change from MOD-UI sync with the LCD. It all seems to work, but I want to review the code better. If for no other reason to understand how to use it. I'm hoping this will eventually allow us to sync plugin bypass events too.

Ask and ye shall receive: https://github.com/sastraxi/pi-stomp/pull/25

I also added some more tests to this branch.

rreichenbach (approved):
Looks good, and seems to work.  I'm really excited to have this.

PR #143, opened 2026-05-12
Refactor: footswitch init and add tests

sastraxi:
Renames `Footswitch.enabled` to `toggled` throughout the codebase. The old name was misleading as it tracked the toggle state of the switch (on/off), not whether the switch itself was enabled.

Disambiguating this frees us up to do the following:
                                                                                               
- `disable: true` already suppresses a footswitch for a given pedalboard; we now save its current value in the Footswitch object
- `longpress` on an encoder now works the same way it already did on footswitches — pedalboard config can override the default callback, and an explicit null clears it. Required threading encoder GPIO switches through a new encoder_switch_map keyed by config id, populated by pistomptre at init.

Tests in tests/v3/test_hardware_config.py cover both the existing footswitch overlay behaviour (longpress set/override/reset/suppress, color persistence) and the new encoder + disable paths.

The pistompcore encoder init still hardcodes a single switch; added a brief XXX comment about it.


rreichenbach (approved):
This looks good and seems to work well.

PR #144, opened 2026-05-12
[WIP] Use git pull to update pedalboards

sastraxi:
Replaces the "Update sample pedalboards" flow (which ran a script that overwrote the sample pedalboards from a fixed remote) with "Sync pedalboards", which pulls/merges from a user-configured git remote. This config value lives in default_config.yml, e.g.

```yaml
pedalboards: https://github.com/TreeFallSound/pi-stomp-pedalboards
```

Local edits are preserved across syncs, to the extent git allows it. Update status (including conflicts) show up on the LCD. This includes:

- "Up to date"  
- "N update(s) applied"                                                                      
- "Sync aborted: conflicts" (with conflicting filenames listed)
- "Sync failed: no network"                                                                  
- "Sync error — see logs: journalctl -u mod-ala-pi-stomp"                                    
- "Pedalboards cloned — restart sound engine to load" (first-time clone)
- "Remote mismatch: local commits not on origin. Resolve via SSH then sync."

### Persistent notification icon

A new orange alert icon appears on the LCD whenever the last sync attempt left a problem (conflicts, network error, remote mismatch, etc.) — so the user knows their pedalboards are out of sync without re-opening the menu. It clears automatically once a sync succeeds.
                                                        
### First-boot behavior
If ~/.pedalboards is missing/empty and a remote is configured, pi-stomp clones it automatically at startup rather than requiring the user to SSH in. This supplements the behaviour provided by pi-gen-pistomp et al.


rreichenbach (commented):
The new Notice is a useful feature.

This isn't quite working as I would expect though. With a clean pi-Stomp install (no pedalboard changes), if I add to my config:
pedalboards: https://github.com/TreeFallSound/pi-stomp-pedalboards

I get the Remote Mismatch Notice.

If you make any changes, it would be helpful to contextualize the notification messages so that the user knows there are regarding Pedalboard syncing.

sastraxi:
I no longer believe this is the right approach. I think we need a larger design discussion around how updates work in general -- pedalboards, user files, LV2 plugins, pi-stomp and mod-ui versions.

PR #146, opened 2026-05-17
Fix RGBA image rendering

sastraxi:
When I introduced a PNG (different branch) that had a yellow transparent background, I saw a yellow box where I expected the icon to be.

We need to provide a mask to the draw call to fix this; without a mask, `paste(src, loc)` does a raw pixel copy, ignoring the alpha channel entirely and overwriting destination pixels with the source's RGB values. This has two effects:

* In test snapshots, semi-transparent fringe pixels got their raw RGB values stamped onto the destination at full opacity, creating halos or bleed around icon edges.

* In production, PIL converted the source to RGB, meaning the background colour came with it at full opacity (pre-composited). Fortunately, `Widget._do_draw` always calls `_draw_erase` before `_draw`, so there's no accumulation risk.


sastraxi:
Already landed in wifi menu branch

PR #148, opened 2026-05-21
Animated wifi connection icon

sastraxi:
Adds a small animation to make it clear that the processing state is an ephemeral one.

* renames `image_path=` to `image=` in ImageWidget; allows creation and replacement with PIL Image in addition to providing a path
* pre-loads animation frames in the LCD
* cycles animation every 2 ticks (UI polls)
* adds test coverage


rreichenbach (approved):
Very Cool

PR #149, opened 2026-05-22
Expand usage of websockets for parameter / BPM setting

sastraxi:
Migrates outbound parameter/bypass/tempo writes from the bespoke `pi_stomp_set` REST endpoint to the existing MOD-UI WebSocket, and adds inbound handling for plugin-bypass echoes so the LCD stays in sync with MOD-UI-originated changes.

This removes our need to maintain a fork that adds the `pi_stomp_set` REST endpoint.

### Outbound messages

- Plugin bypass toggle (non-footswitch) → `param_set /graph/<inst>/:bypass <v>`
- Parameter edit commit → `param_set /graph/<inst>/<sym> <v>`
- Tap tempo → `transport-bpm <bpm>`

### Inbound messages

- `param_set /graph/<inst> :bypass <v>` from MOD-UI is parsed into a `PluginBypassMessage`, updates plugin state, and redraws the LCD plugin row. Idempotent on self-echo.

## Changes

- **Fail-fast startup.** The WebSocket bridge is now required; the previous `try/except` swallow around bridge construction is gone. If the bridge fails to construct or start, handler init raises. Two unit tests pin this.
- **No more rollback on send failure.** The old REST path toggled `plugin.bypass` back if the POST returned non-200. Fire-and-forget is fine here because the WS connection is the source of truth: if we're connected, we're synchronized; if we're not, the bridge will reconnect and the next state change carries us forward.
- **Canonical `instance_id`.** `Plugin.instance_id` is normalized at construction (leading slash stripped). The parser produces bare instance names. Comparisons are plain `==`. `send_parameter` still defensively strips + warns if a caller passes a slashed id (probably best to figure out a way to enforce this at the type layer later)
- Added tests for all the above


rreichenbach (approved):
This is awesome!

PR #153, opened 2026-05-24
Reduce burst latency for encoders

sastraxi:
Instead of using one pending "detent" of an encoder per tick, we now execute all callbacks instantly, passing `n` through (batching) where downstream understands it.

The philosophy: don't draw an intermediate state; just draw the end state if we know we're there. Lowers perceived latency at the expense of smoothness.


rreichenbach (approved):
Seems to work well

PR #157, opened 2026-06-03
Fix scrolling long menus

sastraxi:
Selection navigation would move to an off-viewport widget, but the container failed to scroll, leaving the user with an invisible selection. This fixes this behaviour and adds a number of tests.

I also re-formatted (based on the project-wide ruff rules) and added type hints to the files I touched here; use "Hide Whitespace" for a reasonable review experience:

image


rreichenbach (commented):
I'm not understanding what the problem was or how to repro it.

sastraxi:
> I'm not understanding what the problem was or how to repro it.

If there are too many items in a menu dialog, scrolling past the last one doesn't actually scroll the container to catch up, so it looks like you just lose the selection rectangle. It only happens when all of the branches I was working on are merged together, as the System Menu gets too long for the height of the LCD.

Happy to revert the formatting changes if it makes it easier to review.

rreichenbach (approved):

PR #158, opened 2026-06-04
Strobe tuner + Jack-Client

### LINKED ISSUE #121: Enable on-screen tuner
I found that there is a [tuner plugin](https://pedalboards.mod.audio/plugins/aHR0cDovL2dhcmV1cy5vcmcvb3NzL2x2Mi90dW5hI21vZA==) installed, so I tried adding a switchbox to send signal to the tuner for tuning. 

It looks like this works within the web-ui, but I don't see anything on the pi-stomp itself. 

Is it possible to enable the tuner to display?

sastraxi:
Adds a strobe tuner that can be assigned via footstwitch longpress config or trigged via the System Menu. Closes #121 

### Installing dependencies

On pi-gen-pistomp:

```bash
sudo pip3 install Jack-Client numpy
sudo systemctl restart mod-ala-pi-stomp
```

This will install in the system python environment, which is what the pi-stomp service runs in (3.11).

### Tuning algorithm

I had the LLM implement [this paper](http://audition.ens.fr/adc/pdf/2002_JASA_YIN.pdf), but it wasn't very accurate. Then I realized how it worked and properly tweaked the buffer size / range and it was decent enough for bass. Getting it accurate for guitar, however, took more work.

### Differences from textbook YIN

YIN works one frame at a time: it slides the signal against a delayed copy of itself and looks for the delay where the two line up best — that delay is the period, and you read the pitch off it. The match quality across delays forms a curve with a dip at the true period; standard YIN finds that dip and fits a small parabola to it to pin the period down to a fraction of a sample.

We change a few things:

1. How we read the bottom of the dip. The parabola trick assumes a single clean low point, but on high strings the dip bottoms out flat — two neighbouring delays tie, and the real answer sits between them. A parabola can only point at one of its three samples, so it commits to one side; at those frequencies a single sample of delay is worth ~12 cents, so the reading flips back and forth around the true pitch and never settles. Instead of the parabola we average across the floor of the dip, leaning toward its deepest part, which lands cleanly between samples. (A sharp, unambiguous dip still uses the parabola.)

2. What we do across frames. Plain YIN gives one reading per frame; we report the median of the last five. This is deliberately not smoothing — averaging would drag an occasional octave jump out into a slow slide through every wrong note on the way, whereas a median just throws the bad reading away and keeps the good ones.

3. We add an RMS silence gate and an onset hold-off so the pluck transient doesn't contribute to tuning.


rreichenbach (commented):
I've done the package installs.  I've added tuner enable to footswitch C, but when I longpress it, I get the following:

Traceback (most recent call last):
  File "/home/pistomp/pi-stomp/modalapistomp.py", line 222, in <module>
    main()
  File "/home/pistomp/pi-stomp/modalapistomp.py", line 192, in main
    handler.poll_controls()
  File "/home/pistomp/pi-stomp/modalapi/modhandler.py", line 195, in poll_controls
    self.hardware.poll_controls()
  File "/home/pistomp/pi-stomp/pistomp/hardware.py", line 98, in poll_controls
    s.check_longpress_events()
  File "/home/pistomp/pi-stomp/pistomp/footswitch.py", line 79, in check_longpress_events
    callback()
  File "/home/pistomp/pi-stomp/modalapi/modhandler.py", line 946, in toggle_tuner_enable
    engine.start()
  File "/home/pistomp/pi-stomp/pistomp/tuner/engine.py", line 79, in start
    self._source.start(on_samples=self._ring.write)
  File "/home/pistomp/pi-stomp/pistomp/tuner/source.py", line 41, in start
    self._client = jack.Client(self._client_name, no_start_server=True)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/jack.py", line 216, in __init__
    raise JackOpenError(name, self._status)
jack.JackOpenError: Error initializing "pistomp-tuner-1": <jack.Status 0x11: failure, server_failed>



Jack seems to be running just fine, but apparently not pistomp-tuner.
Any suggestions?  Maybe additional pi-gen/env changes?  jack config?  (eg. JACK_PROMISCUOUS_SERVER?)

We should probably catch that exception instead of crashing.

rreichenbach (commented):
Ah, got it working!

Added two lines to mod-ala-pi-stomp.service

[Service]
Environment=JACK_PROMISCUOUS_SERVER=jack
LimitRTPRIO=95

It's very cool!  Almost too accurate for the high strings (basically always either sharp or flat).

rreichenbach (commented):
A few suggestions to make it less jittery, especially for high strings on guitar:

_IN_TUNE_THRESH  => 5  (wider green range)

IIR_ALPHA = 0.20   (more aggressive smoothing)

self._cents_history: deque[float] = deque(maxlen=5)     (kills more jitter with minimal perceptible lag)

sastraxi:
@rreichenbach https://github.com/TreeFallSound/pi-gen-pistomp/pull/18
I put the max RT priority lower to ensure that the tuner never steals the CPU away from plugins, but you might disagree... 🤔

rreichenbach (approved):
Nice!  Seems much more stable to me for guitar, baritone and bass.

Only other suggestion is to add longpress config for it in default_config.yml on footswtich C.

rreichenbach (approved):

PR #160, opened 2026-06-04
Fix tearing in the emulator LCD

sastraxi:
In the emulator, blit_scaled() could snapshot self.surface while the SPI worker thread was still draining queued blits.

> When the tuner drew "wrap-around" stripes (two update() calls per stripe), the second blit might not yet be on the surface, causing the leftmost stripe to flicker or disappear.

Added a test case as well as a queue for blits that is drained before the pygame blit.


rreichenbach (approved):
I'm fine with this

PR #163, opened 2026-06-05
Fix: plugin bypass states wrong on initial pedalboard load

sastraxi:
Low-priority bugfix.

On boot, pi-Stomp loads plugin bypass values from the LILV-parsed TTL bundle, which reflects the Default snapshot. If MOD is running a different snapshot when the service starts (e.g. because the pi-stomp service crasehd), these values are stale; footswitch LEDs show the wrong state until the user switches presets.

Heres how this happens:

1. set_current_pedalboard() calls
2. bind_current_pedalboard() (LILV defaults) then
3. load_current_presets()

This last method figures out which snapshot is active, but never syncs the actual bypass states back from MOD. preset_change() already does this sync via preset_change_plugin_update(), so the fix is simply to call it on initial load. We add a regression test as well.

### DIFF
--- a/modalapi/mod.py
+++ b/modalapi/mod.py
@@ -582,6 +582,10 @@ class Mod(Handler):
         # Initialize the data
         self.bind_current_pedalboard()
         self.load_current_presets()
+
+        # ensure plugin states (like bypass) are correct on preset change
+        self.preset_change_plugin_update()
+
         self.update_lcd()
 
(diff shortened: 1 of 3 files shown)


sastraxi:
Closing in favour of the branch stack starting at #166. Rather than working around the dual-source-of-truth architecture, that stack addresses it directly.

PR #165, opened 2026-06-06
Set websocket poll to 20hz

sastraxi:
Minimizes the lag time when rapidly changing parameters from pi-stomp, mitigating somewhat but not fully solving the dual-source-of-truth problem.


sastraxi:
Closing in favour of https://github.com/TreeFallSound/pi-stomp/pull/166 -- a more robust stop-gap that puts us on the path to actually fixing the dual-source-of-truth problem.

PR #168, opened 2026-06-06
Track MOD tempo via inbound transport WS message

sastraxi:
mod-ui broadcasts `transport {rolling} {bpb} {bpm} {sync}` on tempo change and in the connect dump. This branch parses it and updates TapTempo unconditionally, so the device tracks MOD tempo whether or not tap-tempo mode is enabled.

Note that there is no live LCD refresh here; the footswitch BPM label refreshes on its normal redraw cycle (if it's visible at all).


rreichenbach (commented):
This one isn't working how I would expect.  But perhaps, just checking out the PR isn't including other prerequisite changes?

If I tap in a new tempo, I don't see it reflected in the MOD-UI unless I do a browser refresh.  This used to work.

And I thought based on the description that MOD-UI tempo changes would be reflected back to the LCD, but I'm not seeing that either.

sastraxi:
@rreichenbach got it -- this was more of an opportunistic sync fix, but it sounds like MIDI tap tempo is actually broken right now. #167 may have fixed it; I'll look into it.

sastraxi:
Fix landed, ready for re-review.

rreichenbach (approved):

PR #169, opened 2026-06-06
Add v1 system fixture + integration test

sastraxi:
Test only changes; adds a fixture for the v1 hardware with the same snapshot functionality and a sanity test.


rreichenbach (approved):
Seems fine.  We probably could abandon v1 at some point but I appreciate that you're including it.

PR #170, opened 2026-06-09
Faster LCD updates on bypass / footswitch changes

sastraxi:
Every plugin-bypass echo and every footswitch press triggered multiple full-panel SPI blits. Now, we only update the plugin (and footswitch) whose state changed via the incoming message.

Included in this is a small fix for footswitch rendering: these were sized to plugin_height, shorter than their drawn halo+label. The old full-panel blit masked the overflow; partial redraws exposed it. Corrected to the intended footswitch_height.

Snapshots unchanged (pixel-identical), confirming the partial path matches the old full-panel output.


rreichenbach (commented):
Hmmm, I'm seeing the almost the opposite with this one.  The footswitch LED and halo update quickly, but the lag for the associated plugin block seems way worse - sometimes almost a second or so.  I also sometimes see an LED flash like it's getting updated twice.

rreichenbach (approved):
Code looks good and seems to work well now

PR #171, opened 2026-06-09
Add configurable LCD SPI speed via system menu

sastraxi:
The ILI9341 controller runs reliably above spec — 56 MHz has been tested stable, cutting full-panel refresh from 80ms to 30ms. This PR makes the SPI clock user-selectable (24/48/56/80 MHz) via System Menu → LCD Speed; the choice is persisted in settings and applied at startup.

The polling divisor in the main loop auto-adjusts so LCD polling tracks somewhat-closely with actual refresh time. Additionally, the ADC SPI speed is set based on the LCD speed; experimental settings set it to the spec (1 MHz) while the base setting maintains the existing 240 kHz value (tested on actual hardware).


rreichenbach (approved):
Seems fine.  I'm seeing footswitch refresh glitches similar as on #170 but this is probably unrelated to this changeset.  I see the same thing regardless of the frequency set.

Would be nice if we had a MessageDialog class that allowed hiding the OK button.  Not a big deal though.

PR #172, opened 2026-06-11
Apply pedalboard bindings immediately

sastraxi:
There's a websocket message (`midi_map`) we can listen to that allows us to update the LCD immediately when we bind footswitches in MOD-UI. This branch does a few things to make us responsive to the signal:

1. Extracts shared controller/parameter/MIDI binding logic to `handler.py`
2. Introduces a `_redraw_after_binding` abstract method so v1 can redraw everything while v3 scopes its change to the footswitch that has been edited
3. Simplifies pedalboard loading using the above primitives
4. Adds v1/v3 snapshot tests to confirm the binding works


rreichenbach (approved):
That's nifty

PR #178, opened 2026-06-13
Footswitch bugfixes

sastraxi:
Reworks the LCD footswitch strip (on v3) so stays in sync with MOD-UI and behaves consistently whether it is bound to a bypass or to an arbitrary parameter.

### What changed

1. A footswitch bound to a parameter other than `:bypass` now derives its on/off state directly from the parameter value, without the bypass inversion. This means toggle parameters like a "Solo" switch light up exactly when MOD-UI says they are on, rather than silently staying constant.
2. Plugin.set_param_value is introduced as the single path for updating a parameter's cached value and pushing that value to every control bound to it. Mod now routes both param_set and connect-dump updates through this helper, so bound footswitches (and future controls) stay in sync without duplicating the update logic.


rreichenbach (approved):
Looks good, works well

PR #181, opened 2026-06-20
v1+v3: don't die on missing last.json

sastraxi:
Some things in the handlers now assume `self.current is not None`. When last.json is unavailable (malformed or missing), modalapistomp now:

1. Writes last.json with pedalboard_list[0].bundle (so future restarts don't repeat this)
2. Calls handler.pedalboard_change(pb) — this is the POST to mod-ui's pedalboard/load_bundle/ endpoint that makes it reload
3. Calls handler.set_current_pedalboard(pb) immediately — this is the critical fix; previously `self.current` stayed None until the file-watch fired 100 ticks later, causing crashes in poll_ws_messages when snapshot WS messages arrived


rreichenbach (approved):
Looks good.  Thanks.

PR #183, opened 2026-06-21
clear /run/lcd.init on each process start

rreichenbach:
always reset and clear /run/lcd.init file on process start so service restarts are reliable.


sastraxi (approved):
Makes sense -- this was admittedly a micro-optimization to get rid of the ~0.7s black flash before the service comes up. Totally fine to get rid of it. Sorry for causing issues on v2 🙏

PR #191, opened 2026-07-02
Footswitch preset fixes

sastraxi:
Clear a footswitch's inherited `midi_CC` when `preset:` is configured: it was letting an unrelated plugin's MIDI-learned binding steal `fs.parameter` and clobber the label.

The label now shows the snapshot name instead of the raw index, and the LCD/physical LED lights for whichever snapshot is currently active.


rreichenbach (commented):
This looks good and seems to mostly work.  One problem noticed though, if you change from a pedalboard with preset/snapshot footswtiches, then change to one without, the function and label for the new pedalboard will "bleed" from the previous.

These two lines added to footswitch.py should fix it:

    def clear_pedalboard_info(self):
        self.toggled = False
        self.disabled = False
        self.display_label = None
        self.set_category(None)
        self.preset_callback = None
        self.preset_callback_arg = None     # cuz it's used as a condition too
        self.parameter = None           # not sure it's necessary but seems like good hygene
        self.clear_relays()

rreichenbach (approved):

PR #197, opened 2026-07-10
Optimize parameter drawing; give nav controllers the same behaviour as tweak

sastraxi:
* we now draw both filled and unfilled parameter graphs once and then blit only the dirty part
* volume now gets 128 steps, just like MIDI, instead of 256 (might want to undo this)
* the nav encoder now uses acceleration, but only in the parameter menu -- it now acts the same way as tweak


rreichenbach (approved):
That feels nice.  Acceleration when you need it, fine tuning when you don't.

sastraxi:
@rreichenbach what do you think about the # of detents on the volume? Would you like more precision or does the speed feel good?

PR #198, opened 2026-07-10
[WIP] Put ADC inputs to sleep when they are disconnected

sastraxi:
Introduces a state machine for each analog MIDI control. When we are receiving pure noise from a disconnected ADC input, we now transition it to `ASLEEP`, waking it up (`-> AWAKE`) when movement is detected again. This is a live determination, so pedals can still be plugged in / unplugged during normal operation.

Asleep inputs do not re-paint the UI, nor do they send MIDI events.

### Checklist

- [ ] fix the larger noise jumps -- it might be as easy as stddev over the last N samples with really close mean?


sastraxi:
I think the cleaner solution is to allow selecting the Analog w_control widget (like we did the footswitches row) and toggle-disable it. We can default to the original config -- should branch off https://github.com/TreeFallSound/pi-stomp/pull/239 though

PR #199, opened 2026-07-10
[WIP] Refactor to full input sink architecture

sastraxi:
We did a half-measure to get to the `InputSink` standard, leaving rotary encoders alone (there was already an implementation tied tightly to the UI). This re-uses the same architecture to drive selection to reduce "special cases" in the code, and brings in some more regression tests.


sastraxi:
Superseded by #203

PR #201, opened 2026-07-11
Optimize the render loop

sastraxi:
A few things were going wrong to make rendering a lot slower than it should have been:

* `tint_mask` turned into a general-purpose, caching helper for alpha painting
* a bunch of places refactored to use tint_mask
* rounded rects are now cached
* arc ring rewritten to be more efficient + cached
* remove `font.font_metrics` and instead cache `get_text_size` / `get_text_bbox`
* cache text drwaing
* add batching so immediate draws can't race coalesced draws
* coalesce new draws into pending pushes

As a result, the v2 is now much more usable and will have more CPU available for effects. However, the caches will use about 5-6 mb more RAM.

The snapshot changes are all small pixel alpha blending rounding changes; nothing visible has changed.

Finally, I removed all profiling code (was from a previous investigation and was never cleaned up).


rreichenbach (approved):
Looks good.

PR #203, opened 2026-07-12
[WIP] Declarative UI bindings

sastraxi:
Right now, tweak1/2/3 behaviour can be overridden using `InputSink`. This is fine and works on v3, but has some issues:

1. Learnability. How does a user actually learn that e.g. in a parametric EQ, tweak3 gets overriden from "volume" to "tweak Q of selected band"?
2. Portability. v2 is a fully customizable version, which could have any number of potentiometers for control. It could even have zero!

### Learnability: Displaying bound parameters

This refactor sets the stage to show, generically, what hardware controls are mapped to which parameters. These will show contextually as badges (A/B/C/D/1/2/3) beside whatever the UI is showing the parameter being tweaked:

1. On fullscreen plugin panels, the status bar shows e.g. ① beside band decibels, ② beside frequency, etc.
2. In parameter menus, Ⓐ is shown beside the parameter (e.g. bypass) that is currently associated with that footswitch, ① beside the parameter that tweak1 is currently bound to, etc.

### Portability: Supporting one NAV encoder everywhere

The new plugin panels override the tweak encoders to make it easier to tweak your sound live:

* tweak1 becomes a universal "change the value under the selection cursor"
* parametric eq extends this with tweak2 = frequency, tweak3 = Q (inv. bandwidth)
* the compressor panel also uses these to provide threshold + ratio

However useful these are, they're quite undiscoverable and likely will be frustrating to new users. What will be even _more_ frustrating is when they cannot figure out how to actually change any values because they're not used to these tweak encoders getting overridden!

Aside from the hints (e.g. seeing "① 1.11 kHz"), we're refactoring every panel to support the standard plugin parameter flow that we've always had: long-pressing a parameter (whatever is representing it: an EQ handle, an arc ring, a dropdown, or a menu item) will always let the user edit the parameter with the NAV by bringing up a menu:

* this usually means just the raw parameter dialog; user-closeable.
* parametric eq is the exception: it will show the "band parameters" in a sub-menu: Gain, Frequency, Q/Bandwidth, and Bypass. As with other menus, the tweak encoders bound to these will be shown beside the parameter names, which becomes another way users learn.

This means we'll drop the long press = reset to pedalboard-saved flow. All plugin panels have a "Reset all" button, anyway.


sastraxi:
Going to merge this in a different way.

PR #204, opened 2026-07-12
Make EQ curves representative of underlying filters; calculate Q if plugin uses bandwidth

sastraxi:
* different eq filters require different visualization; annotate at `BandSpec` level
* split filters out into their own file and implement regalia_mitra
* fix fil4 EQs and others that use regalia_mitra filters for bands
* alway show "Q", but derive it from bandwidth (1/bw) if that's what the plugin uses


maarthome:
Hi Cameron...

I almost feel like I am stalking you..... 😅😅😅
Hardly dare to start about the filters again....

The Bandwidth Plot around the selected frequency gets smaller when Q is increasing, so great! .... but... To increase Q, the user has to turn the knob counter clockwise.... which feels counter intuitive to me....

Just saying... 

Cheers,
Maarten

PS Thanks again for the WONDERFUL improvements of the pistomp software!!  
PPS I love the update possibility!!! Although I was a little hesitant to select recovery mode..... Perhaps naming it "Update and Recovery" ?

sastraxi:
@maarthome please keep the feedback coming. It's really rewarding when someone uses the software you're building. Regarding Q -- you're right, I'm messing around with some VSTs and they indeed act the way you're saying (clockwise to increase Q --> decrease bandwidth). This whole thing has happened because the fil4 eq calls its parameter "q" but it's actually a bandwidth (octaves) value 🫠

I think "Updates and Recovery" is a great call-out. I'll put it on the list.

sastraxi:
@maarthome try this one when you have a moment.

maarthome:
Hi Cameron,

That is very fassssst!! Now indeed Q becomes bigger whem turning clockwise.

But I think something happened to the calculation of the Q value? The plugin indicates 0.06 for BW (rounded I guess) and the Q shown on the LCD screen is 23.08..... So not 1/BW... (= 16.67)

Just saying...
Cheers,
Maarten

sastraxi:
Maarten,

I looked it up and the Q value is not always 1/BW. I used the formula I found online to do the conversion.

rreichenbach (approved):

PR #206, opened 2026-07-14
Speed up Model pi3a+ SPI speeds for coalescing decisions

sastraxi:
We have a cost model to predict how long an SPI write will take, so we always choose the fastest path for whatever the pending paint is on the next frame. This allows us to draw small things very quickly (50hz) while preventing falling behind  for large writes.

This model was trained on a Pi 5 and applied everywhere. However, v2's Pi 3A+ per-pixel cost is 6.2x v3's. As such, we now now read the SoC from `/proc/device-tree/compatible` and use sensible decisions.

In the future, we should investigate putting the SPI pushes on their own thread entirely.

### Bonus work

Replace the numpy 565 pack with an SDL convert-blit into a 16-bit staging surface, and stop forcing PanelStack's root to RGBA when dimming: the root is a blend destination, and an SRCALPHA source sends the LCD's convert-blit down SDL's per-pixel alpha-blending path (~7x slower, on every push).

Adds tools/bench_adc_contention.py, extends bench_lcd_device/bench_pack_variants for refitting, and hardens deploy.sh (excludes .venv/caches, reports service start failure).


rreichenbach (approved):
Seems to work well on v2 and v3

PR #207, opened 2026-07-14
Improve startup performance: drop lilv + requests, defer pedalboard hydration

sastraxi:
The pistomp core (v2; rpi 3a+) was taking almost 7 seconds to load.

* Pedalboard parsed every board's .ttl through lilv at construction — a World per Pedalboard, bundle load, plugin scan — to fill in plugins and connections we only read once a board is actually selected. Boards now construct bare and carry a `hydrated` flag; the plugin/connection data is pulled from mod-ui's HTTP API on demand.

* This is actually kind of big: we don't actually need lilv, because we can get everything we need from data we're already loading from MOD-UI.

* We also move away from `requests` to save 1.3s of startup time(!) on this device. `pistomp/httpclient.py` implements the minimum viable amount of the requests API via urllib instead.

* Some smaller ones: the NAM panel scanned the model wav to derive gain; hardcode it instead. And the tuner imports at first use rather than at handler construction.

At the end of this it's down to around ~3.5-4s.


rreichenbach (approved):
Pretty cool that we can leave all the lilv to MOD.

PR #212, opened 2026-07-18
Support extra_data on dynamically-added plugins

sastraxi:
### Problem

A plugin added from the mod-ui web UI mid-session was missing `extra_data`, so NAM tiles show a generic name instead of the model, and the Notes panel opens blank. Boards loaded from disk are unaffected — `hydrate` has the instance number and reads the TTL.

### Mechanism

`plugins.customization.lookup` populates `extra_data` by reading `{bundlepath}/effect-{instance_number}/effect.ttl`. `Pedalboard._build_plugin` (the `add` WS path) has neither the bundle nor the number, so it calls the customizer with defaults and the parser never runs.

## Fix: consume patch_set

mod-ui already broadcasts what the TTL parsers are digging for:
```
patch_set {instance} {writable} {paramuri} {valuetype} {valuedata}
```
... whenever a writable property changes, and it also replays every stored property on the connect dump. The `paramuri`s are the same ones `_MODEL_RE` and `_NOTES_RE` match in the TTL — `…neural-amp-modeler-lv2#model`, `…notes#text` — because that is how patch properties serialize into `effect-N/effect.ttl` in the first place.

The fix is to respond to these extra fields:

- `ws_protocol.py` — add a `PatchSetMessage(instance, param_uri, value_type, value)` case
- `plugins/customization.py` — add a second registration hook alongside `extra_data_fn`, keyed by property URI, that builds `PluginExtraData` from a live value rather than TTL text.
- `modhandler.py` — on `PatchSetMessage`, resolve the plugin, rebuild its `customization.extra_data`, then repaint.

This covers the case the TTL never can: a *freshly* added plugin has no `effect-N` directory at all until the board is saved. It also makes extra_data track mid-session changes — swap the NAM model in the browser and the tile follows.

## Bonus: the Notes TTL regex

`_NOTES_RE` required `"""triple quotes"""`. Turtle only long-quotes when the text needs it, so any single-line note serialized as `"Abc123"` and silently failed to parse — blank panel, no pencil, on the *loaded-board* path where extra_data was supposed to work. Fixed alongside; unrelated to the dynamic-add gap above.


rreichenbach (approved):
Very cool!

Issue #215, opened 2026-07-22
MIDI Bpm should be more accurate ; 2bpm per encoder tick not ideal


jpremill:
based off the feedback from the closed PR for #213  I am working on this change

PR #216, opened 2026-07-22
set bpm from encoder via ws

jpremill:
based on feedback to address #215  based on #213  conversation. I tried to keep it more atomic than I am used to; 
I have only verified via the the tests on these and have not tested on. physical the device yet.


modalapi/modhandler.py
```diff
@@ -418,14 +419,20 @@ def _handle_encoder(self, event: EncoderEvent) -> bool:
...
 
         # Unconditional, and must stay that way: an unbound encoder has no row,
         # and this emit is the only way mod-ui sees its CC to MIDI-learn it.
         # Emission is hardware-level, below the table (see input/README.md).
-        self._emit_midi(c, emit_value)
+        # Transport BPM parameters bypass 7-bit MIDI CC emission for high-precision WebSocket transport.
+        if c.parameter is None or not (
+            c.parameter.instance_id == Pedalboard.TRANSPORT_INSTANCE_ID and c.parameter.symbol == BPM_SYMBOL
+        ):
+            self._emit_midi(c, emit_value)
```
sastraxi: Parameters are reactive (subscription based), so I wonder if instead of detecting at action-time we can proactively change their subscription instead. I think it would be easier to maintain.

modalapi/modhandler.py
```diff
@@ -869,6 +876,18 @@ def _handle_ws_message(self, msg: WebSocketMessage):
...
+            # MIDI slave, another HMI). :rolling's enum flips Playing/Stopped.
+            if self._current is not None:
+                tp = self.current.pedalboard.transport_plugin
+                if tp is not None:
+                    if ROLLING_SYMBOL in tp.parameters:
+                        tp.set_param_value(ROLLING_SYMBOL, 1.0 if msg.rolling else 0.0)
+                    if BPM_SYMBOL in tp.parameters:
+                        tp.set_param_value(BPM_SYMBOL, msg.bpm)
+                    if BPB_SYMBOL in tp.parameters:
+                        tp.set_param_value(BPB_SYMBOL, msg.beats_per_bar)
```
sastraxi: I think we can skip a lot of these nullability checks. I don't see a reason to ever not instantiate a transport plugin; every pedalboard is "playable" even if the user never uses that feature. And when would we not want to store those parameters? Basically, we should make setting `tp` params always safe.

I know this part of the design might be inherited... I still wonder if we can do better.

modalapi/pedalboard.py
```diff
@@ -181,8 +234,73 @@ def hydrate(self, plugin_dict) -> None:
...
 
+    def _build_transport_plugin(self, time_info: dict | None) -> Plugin.Plugin | None:
+        """The /pedalboard pseudo-instance carrying :bpm/:bpb/:rolling. Built
+        from mod-ui's timeInfo block. None when the board reports no transport
+        metadata (available == 0 or absent)."""
+        if not time_info:
+            return None
+        available = int(time_info.get("available", 0) or 0)
+        if available == 0:
+            return None
```
sastraxi: I would like to tighten the return type to `Plugin.Plugin` and create a stub transport plugin if the pedalboard itself doesn't have one. MOD-UI's
```
transport {rolling} {bpb} {bpm} {sync}
```
message is broadcast unconditionally, regardless of how a board is saved. It saves -1 for each unmapped param, and if it for some reason doesn't exist, we also know it's unmapped. But the parameters are still there. They'll just be unbound if the struct doesn't exist.

IMO we should a) unconditionally create the plugin and b) unconditionally create each parameter. Pretty sure the bitmask isn't useful for us. WDYT?

jpremill: I think I don't have enough time in the codebase to really have an opinion so I took your suggest as gospel.

rreichenbach (approved):
Since this is more his domain, I'll defer to Cam on the code, but I did test this on hardware and it works just as I would expect.  The control maps fine, and changing the setting tracks whether from Encoder, Footswitch or MOD-UI.

modalapi/pedalboard.py
sastraxi: Coming back to this, I'm still pretty convinced. So let's try it and we can re-evaluate if problems arise.

modalapi/modhandler.py
```diff
@@ -426,7 +430,9 @@ def _handle_encoder(self, event: EncoderEvent) -> bool:
         # Unconditional, and must stay that way: an unbound encoder has no row,
         # and this emit is the only way mod-ui sees its CC to MIDI-learn it.
         # Emission is hardware-level, below the table (see input/README.md).
-        self._emit_midi(c, emit_value)
+        # Transport parameters bypass 7-bit MIDI CC emission for high-precision WebSocket transport.
+        if c.parameter is None or c.parameter.instance_id != Pedalboard.TRANSPORT_INSTANCE_ID:
```
sastraxi: We need to still send MIDI for beats-per-bar and rolling; this should look at the symol

modalapi/modhandler.py
```diff
@@ -1442,7 +1460,7 @@ def parameter_value_commit(self, param, value):
                     self._emit_midi(controller, int(value))
                 return
 
-        if not self._is_pedalboard_loading:
+        if not self._is_pedalboard_loading and param.instance_id is not None and param.instance_id != Pedalboard.TRANSPORT_INSTANCE_ID:
```
sastraxi: Same as above (branch on symbol not instance ID)

modalapi/modhandler.py
```diff
@@ -1427,7 +1446,6 @@ def parameter_value_commit(self, param, value):
         # Audio parameter (volume, EQ, etc.) - handled locally, no remote update needed
         if param.instance_id is None:
             self.audio_parameter_commit(param.symbol, value)
-            return
```
sastraxi, line 1430: We need this early return otherwise the audio parameters will emit CCs below.

modalapi/modhandler.py
```diff
@@ -1798,8 +1816,11 @@ def audio_parameter_commit(self, symbol, value):
     def get_callback(self, callback_name):
         return util.DICT_GET(self.callbacks, callback_name)
 
-    def set_mod_tap_tempo(self, bpm):
+    def set_mod_tap_tempo(self, bpm: float | None) -> None:
         if bpm is not None:
+            self._last_bpm_change_time = time.time()
```
sastraxi: I think maybe left over from an earlier design?

modalapi/modhandler.py
```diff
@@ -1798,8 +1816,11 @@ def audio_parameter_commit(self, symbol, value):
...
+            if self.ws_bridge is not None:
+                self.ws_bridge.send_bpm(bpm)
```
sastraxi: Why do we have to send this over WS? Was the REST laggy? Regardless we need to put `_rest_post` in the else clause to avoid a double-send. But I'd prefer to just do it one way or the other.

tests/v3/test_transport_bindings.py
```diff
@@ -197,3 +290,252 @@ def test_load_time_binding_labels_on_load(v3_system: SystemFixture, make_plugin,
...
+    # Inject external transport frame with decimal BPM (120.5)
+    ws_bridge.inject("transport 1 4.0 120.5 link")
+    handler.poll_ws_messages()
+
+    assert tp.parameters[BPM_SYMBOL].value == 120.5
+
+
+def test_encoder_bpm_turn_without_websocket_bridge_falls_back_to_rest_post(
+    v3_system: SystemFixture, make_plugin
+):
```
sastraxi: This test could be deleted once we decide REST vs WS

sastraxi (changes requested):
This is close! A couple real bugs here worth taking a look at.

It would be helpful if you could test on device. Personally, I like to use `./deploy.sh` to just copy local files over to test, but you might prefer @rreichenbach 's approach which is to edit files directly on the device. There's a helper you can use to turn the deployed source code back into a git repository if it helps:

https://sastraxi.github.io/pistomp-manual/developers/getting-started/#dev-mode-git-managed-source-on-the-device

tests/v3/test_transport_bindings.py
sastraxi: @jpremill would still suggest removing this. The service is no longer considered to be healthy if the WS bridge is not up.

sastraxi (approved):
Good to merge after the final comments are addressed!

modalapi/modhandler.py
sastraxi: @jpremill thoughts?

jpremill:
@sastraxi  I've been strapped for time my bad

sastraxi:
> @sastraxi I've been strapped for time my bad

@jpremill there's no need to apologize! Thanks so much for the contribution. This will be in the next release.

PR #218, opened 2026-07-23
217 midi unlearning tweak12 bug / address footswitch complete teardown

### LINKED ISSUE #217: Tweak1/2 control bar displays are stale after clearing a MIDI binding
### Steps to reproduce

1. Open up the pedalboard in http://pistomp.local
2. MIDI learn a parameter to Tweak1/2
3. Note that the display now shows the value and you can select it. Everything is in sync
4. Now remove the MIDI mapping in MOD-UI

### Expected

The display of the control should go back to being grey with "none" text; the parameter should now be unbound and no longer show the parameter dialog when rotated. Fixed when you go to another pedalboard and back to this one

### Actual

The control appears to be bound, and continues sending the MIDI messages. The parameter value drifts from the real one in MOD, which is correctly unchanged.

jpremill:
fixes #217  
I addressed the early return in __apply_midi_binding_  that allowed unmapping to reach _old_controller.unbind_from_parameter()_, which exposed that footswitch unmapping was incomplete — _Footswitch_ was missing an _unbind_from_parameter()_ override to clear its parameter label/category, and _update_footswitch()_ was missing an else: branch to repaint unmapped switches back to grey.I noticed this after the first commit.  I was hesitant to continue the fix just based on needing to update the footswitch too, but I suppose this is the whole point of the testing.


jpremill:
@sastraxi  I still haven't setup the deployment to the hardware device,so again this is all based on the reasonable assumption of the tests passing.  The test snapshots were very helpful to identify issues, but I wasn't able to manually test using the emulator. I suspect it has something to do with the MIDI registration on mac silicon.

sastraxi:
@jpremill I gave some tips on how to deploy in the other PR! Will be reviewing this one shortly.

By the way, I hope you don't mind, I had to change the upstream branch due to some shuffling around on my end. There's a small merge conflict now.

pistomp/handler.py
```diff
@@ -206,40 +205,44 @@ def show_fullscreen_panel(self, plugin, panel_cls) -> None:
...
-            param.set_binding_range(binding_range)
-        if param.binding == binding:
+        if param is None or param.binding == binding:
             return
+
+        old_binding = param.binding
         controller = self.hardware.controllers.get(binding)
+        old_controller = self.hardware.controllers.get(old_binding) if old_binding is not None else None
+
+        if old_controller is not None and old_binding != binding:
```
sastraxi, line 236: This should be fine after the addition of `binding_range` to fix a bug with the "Advanced" binding features of MOD-UI (we were ignoring the parameter range before), but it would be good to have a test that explicitly makes sure a different range bound on the same parameter still works as we expect it to. Sorry for the churn.

jpremill:
@sastraxi  I saw the guidance; its just a matter of making the time to do it. I just have been finding a few minutes in the evening to make some progress; I think after this changeset, i'll get the env setup before pushing changes.

sastraxi:
@jpremill awesome! No rush. I'm going to be away for a week starting Sunday, so there may be some delays from my end in review.

sastraxi:
@jpremill let me know what the next steps are here. No huge rush but I want to start prepping for a release and ideally we get these contributions in!

jpremill:
@sastraxi its up to you-  I did just do a push to handle  controller teardown and added some tests

jpremill:
@sastraxi  slightly unrelated, but do you have a roadmap or separate board for feature planning? seems like this project is ripe for that.

sastraxi:
> @sastraxi slightly unrelated, but do you have a roadmap or separate board for feature planning? seems like this project is ripe for that.

We don't have a public one right now. Currently I'm looking into things like BLE-MIDI support and how to make MOD-UI a little more user-friendly / deeply integrated with pi-Stomp. If there's anything you'd like to suggest, open up an issue and let's discuss!

PR #219, opened 2026-07-24
[WIP] Parameter sinks

sastraxi:
There are a few different ways of doing parameter reactivity... recent changes in this area made me think we're not going far enough in the subscription model. This is an attempt to fix that. I'm not sure yet if it's the right approach.


sastraxi:
Already merged into `main`.

PR #223, opened 2026-08-07
Fix longpress chords firing every member's solo action

sastraxi:
**Bug:** the singleton grace was timed from the press timestamp, but detectors deliver a longpress 0.5s after the press while the grace was 0.4s — expired on arrival, so the chord never won the race.

**Fix:** resolve synchronously. When the first member reaches the longpress time, every other member is already held, so the chord is decided there. Both gestures fire at 0.5s; no added latency.

### Changed

- **Pedalboards now default to A/B: previous pedalboard + C/D: next pedalboard.**
- footswitch_chords.py — groups hold members (via LongpressMember Protocol), not a count. Chord needs all members PRESSED (unspent); firing marks the group spent so remaining members stay silent. WINDOW, timestamps and tick() deleted; poll() only re-arms on release.
- gpioswitch.py — gains the RELEASED/PRESSED/LONGPRESSED field AnalogSwitch already had, incl. clearing on release after a delivered longpress.
- footswitch.py — press_state property; reads RELEASED when disabled (polling stops, a frozen PRESSED is a phantom chord partner).
- handler.py / modhandler.py — _tick_chords → _fire_longpress_groups(fs), called from longpress dispatch instead of per poll.
- hardware.py — register(fs) not register(fs.longpress_groups).
- config.py — next_pedalboard/previous_pedalboard added to the longpress enum (in the handler's callback map all along, never in the schema).
- Unknown group names now warn instead of vanishing silently.
- input/README.md — the "timing-deferred" justification no longer applies.


rreichenbach (approved):
Nice job getting that to work.

PR #227, opened 2026-08-09
Tweak footswitch UI and add a longpress bindings dialog

sastraxi:
* make the footswitch widget selectable
* remove `set_mod_tap_tempo` from the last of valid longpress actions (it's the actual shortpress tap)
* fix snapshot binding colours (use default; makes them more visible)
* nudge footswitch text up 1px (makes it more visible, shows a bit of descenders)
* fix pedalboard over-scrolling: now the canvas only moves if the new item is not completely visible
* adds a dialog that shows the current pedalboard bindings, available via shortpress/longpress on the footswitch widget
* remove a superfluous test and add new ones


rreichenbach (approved):
Hot Damn.  Very clever and elegant solution.

PR #231, opened 2026-08-10
feat: add clear_binding_range support and update unmap logic to resto…

### LINKED ISSUE #226: Unmapping a parameter leaves it clamped to the old MIDI sub-range
Follow-up from the #217 fix (PR #218). A parameter that was MIDI-mapped to a narrowed sub-range keeps that narrowed range after the mapping is removed, with no way back short of a pedalboard reload.

### Steps to reproduce

1. MIDI-learn a parameter to an encoder with a custom sub-range — e.g. a 0.0–1.0 port addressed as 0.1–0.2.
2. Remove the mapping in MOD-UI.

### Actual

`param.minimum/maximum` stay at `(0.1, 0.2)`. The parameter is now unbound but still clamped to the sub-range a since-deleted mapping imposed, so the parameter dialog paints the narrowed endpoints and the value can't reach the plugin's declared extents.

### Expected

Unmapping restores the plugin's **declared LV2 range**.

### Why it isn't just "apply the unmap frame's range"

The unmap frame carries a default `0.0 1.0`, which is not the declared range — stamping it onto, say, a 30–800 Hz port would be worse than leaving the sub-range in place. PR #218 deliberately guards against that (`test_v3_midi_unlearn_preserves_sub_range` pins the behaviour), and that guard is an improvement on the previous unconditional apply. This issue is about finishing the job.

### Blocker

`Parameter` doesn't retain the declared range. `common/parameter.py:137` computes `declared_minimum` / `declared_maximum` as locals in `__init__` and folds them straight into the effective `self.minimum` / `self.maximum`; `set_binding_range` then overwrites those. Restoring on unmap requires keeping the declared pair as fields first.

jpremill:
should fix #226 

### Problem

When a parameter is MIDI-learned to a narrowed sub-range (e.g. a `0.0–1.0` port addressed as `0.1–0.2`), removing the mapping in MOD-UI never restores the plugin's declared LV2 range. The parameter stays clamped to the dead sub-range until a pedalboard reload, so the LCD dialog paints the wrong endpoints and the value can't reach the plugin's declared extents.

### Root cause

`Parameter.__init__` folded the plugin's declared LV2 range into `self.minimum` / `self.maximum` as **local variables** and discarded them. Once `set_binding_range()` narrowed the effective extents, the original declared pair was unrecoverable — unmap had nothing to restore to.

PR #218 guarded against the unmap frame's default `0.0–1.0` overwriting non-normalised ports (correct), but it left the parameter permanently stuck at the sub-range.

### Fix

1. **`common/parameter.py`** — keep the declared LV2 range as fields (`declared_minimum` / `declared_maximum`) instead of tossing them; add `clear_binding_range()` to restore them. Both it and `set_binding_range()` now clamp `_value`, guard against idempotent replays, and notify observers so active dialogs repaint immediately.
2. **`pistomp/handler.py`** — unmap is now detected by the MOD-UI `-1:-1` sentinel rather than `controller is None`. That disambiguates a real unmap (restore declared range) from an external/free MIDI CC (preserve its genuine sub-range), which the old guard had wrongly conflated.

### Tests

- `test_v3_midi_unlearn_restores_declared_range` — unmap restores the declared LV2 range.
- `test_v3_midi_learn_free_cc_preserves_sub_range` — a free/external CC keeps its real sub-range (locks the sentinel behavior).
- `test_clear_binding_range_restores_declared_range` + `test_binding_range_notifies_subscribers_and_clamps_value` — unit coverage.
- `test_set_binding_range_is_idempotent` / `test_clear_binding_range_is_idempotent` — replays produce zero notifications.

### DIFF
--- a/pistomp/handler.py
+++ b/pistomp/handler.py
@@ -224,9 +224,12 @@ class Handler(InputSink):
         controller = self.hardware.controllers.get(binding)
         # The range can change without the binding (re-address the same CC to a
         # different sub-range), so apply it before the binding-unchanged bail.
-        # Never on unmap: -1:-1 names no controller and must not reset the range.
-        if binding_range is not None and controller is not None:
+        # On unmap (binding "-1:-1" or "-1"), restore the plugin's declared LV2 range.
+        is_unmapped = binding in ("-1:-1", "-1")
+        if binding_range is not None and not is_unmapped:
             param.set_binding_range(binding_range)
+        elif is_unmapped:
+            param.clear_binding_range()
         if param.binding == binding:
             return
 
--- a/tests/test_parameter_binding_range.py
+++ b/tests/test_parameter_binding_range.py
@@ -61,6 +61,52 @@ def test_set_binding_range_preserves_identity():
     assert id(p) == before
 
 
+def test_clear_binding_range_restores_declared_range():
+    """Calling clear_binding_range resets minimum and maximum to declared_minimum and declared_maximum."""
+    p = Parameter(_port(30.0, 800.0), 100.0, binding="0:70", binding_range=(100.0, 400.0))
+    assert (p.minimum, p.maximum) == (100.0, 400.0)
+    p.clear_binding_range()
+    assert (p.minimum, p.maximum) == (30.0, 800.0)
+
+
+def test_binding_range_notifies_subscribers_and_clamps_value():
+    """set_binding_range and clear_binding_range notify observers and clamp value if out of bounds."""
+    p = Parameter(_port(0.0, 1.0), 0.9, binding="0:70", binding_range=(0.0, 1.0))
+    notifications = []
+    p.subscribe(lambda param: notifications.append(param.value))
+
+    # Narrow range past current value (0.9 -> 0.5 max)
+    p.set_binding_range((0.0, 0.5))
+    assert p.value == 0.5
+    assert len(notifications) == 1
+
+    # Clear binding range back to 0.0 .. 1.0
+    p.clear_binding_range()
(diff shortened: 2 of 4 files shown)


pistomp/handler.py
```diff
@@ -224,9 +224,12 @@ def _apply_midi_binding(
         controller = self.hardware.controllers.get(binding)
         # The range can change without the binding (re-address the same CC to a
         # different sub-range), so apply it before the binding-unchanged bail.
-        # Never on unmap: -1:-1 names no controller and must not reset the range.
-        if binding_range is not None and controller is not None:
+        # On unmap (binding "-1:-1" or "-1"), restore the plugin's declared LV2 range.
+        is_unmapped = binding in ("-1:-1", "-1")
```
sastraxi, line 228: nit: let's pull this up to a constant that tells future devs where these values come from (MOD-UI? mod-host?) and what they mean

tests/test_parameter_binding_range.py
```diff
@@ -61,6 +61,52 @@ def test_set_binding_range_preserves_identity():
     assert id(p) == before
 
 
+def test_clear_binding_range_restores_declared_range():
+    """Calling clear_binding_range resets minimum and maximum to declared_minimum and declared_maximum."""
```
sastraxi, line 65: [nit] I think these comments aren't super necessary; they basically repeat what the test names are

sastraxi:
@jpremill looks good to me, I'll give it a test later tonight.

rreichenbach (approved):
That works.  Didn't realize til I tried it was an issue.

PR #232, opened 2026-08-12
Backup/restore progress UI

sastraxi:
### Progress bar
Created a reusable `uilib/progress_bar.py` from the NAM capture panel. Per-instance colour stops (HEAT_STOPS for NAM's fixed-duration capture, STEEL_STOPS for archiving) with matching label pairs, and tweaked padding / tick count to keep things looking sharp.

### Backup / restore features
* `modalapi/archive.py` runs the (un)archive subprocess off of the UI thread with size-weighted progress
* `ui/archive_panel.py` shows a fullscreen modal with the archive progress visualized using the new progress bar.
* atomicity: instead of re-adding to the same archive, `util/data-backup.sh` builds a new one and swaps it in on success
* backups are now cancellable; restores remain uncancellable

### Fixed
* no longer look at mtime to determine whether or not to restore individual files; always restore (removes `-u`)
* the archive is no longer purely additive; e.g. deleted pedalboards are deleted from the backup
* no longer uses `sudo` (all files are pistomp-owned) so that we can cancel the subprocess if desired
* button now vertically centres its label properly


rreichenbach (approved):
Looks good.

PR #237, opened 2026-08-14
Fix pedalboard refetch / last.json restore race condition

sastraxi:
Because `last.json` et al are part of the restore dump, there is a race condition where if the mtime is re-scanning during restore, pi-stomp can crash. This PR gates it on a `_restoring` flag -- this flag only ever transitions from `False -> True` because the last step of restore is to restart the service.

Also fixes a bug where the pedalboards list wasn't correctly being emptied when re-scanning MOD; this meant deleted pedalboards would still show up in the selector.

Bumps to v3.3.1.


rreichenbach (approved):
Looks good.  With OTA now, restoring from USB should be less common.

PR #238, opened 2026-08-15
Reconcile local sinks on successful CC and ALSA emission

jpremill:
## Summary                                                                                                                        
Fixes an issue where editing a parameter via `Parameterdialog` (or local hardware controls) causes the value to snap back to its initial boot/preset setting on subsequent events. I encountered this while goofing around with the settings.
                                                                          
### Root Cause                                                                                                                    
Under the strict 3-motion parameter model (`preview`, `reconcile`, `commit`), edits published through `commit(value, sink)`  remain in a pending state until a `reconcile()` confirms the change.                                                                
While standard plugin parameters receive an asynchronous `param_set` WebSocket echo from MOD-UI, **local non-echoing sinks**      (`_publish_audio` for soundcard ALSA writes and `_publish_cc` for hardware-bound MIDI CCs) receive no remote echo from `mod-host` or MOD-UI. Because `param.reconcile()` was never called for these sinks, `_confirmed` was permanently stuck at the initial value, causing any subsequent touch, expression input, or sub-range recalculation to roll the parameter back to its boot-time value.       

### Changes                                                                                                                       
1. **`modalapi/modhandler.py`**:                                                                                                  
       - Explicitly reconcile `param.reconcile(param.value)` within `_publish_audio` and `_publish_cc` once the write / CC emission   
  succeeds.                                                                                                                           
2. **`tests/v3/test_parameter_edit_rollback.py`**:                                                                                
       - Added TDD regression tests verifying NAV encoder step updates in `Parameterdialog`, safety rollback when                     
  `_is_pedalboard_loading` is active, and confirmation for CC-bound parameters.                                                       
3. **`CHANGELOG.md`**:                                                                                                            
       - Added changelog entry under `### Fixed`.                                                                                     
### Verification
 - `uv run pytest tests/v3/test_parameter_edit_rollback.py` passes (3/3).                                                          
 -  Full test suite passes: 1,464 passed, 2 skipped in ~31s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Global EQ now accounts for the current sample rate
 
 ### Fixed
+- Fix parameter rollback on MIDI CC and ALSA local sinks by reconciling confirmed state upon successful emission
 - WiFi: fixed WPA3 support
 - Fix footswitch ↔ plugin param binding updates in menus
 - Fix inverted Q factor in x42 EQ panel
--- a/modalapi/modhandler.py
+++ b/modalapi/modhandler.py
@@ -1237,11 +1237,13 @@ class Modhandler(Handler):
     def _publish_audio(self, param: Parameter) -> bool:
         """ A local ALSA write. No remote echo, so the send always lands."""
         self.audio_parameter_commit(param.symbol, param.value)
+        param.reconcile(param.value)
         return True
 
     def _publish_cc(self, controller: EncoderController, param: Parameter) -> bool:
         """Publish the parameter (to MOD-UI or anything else) via MIDI CC."""
         self._emit_midi(controller, controller.to_midi(param.value))
+        param.reconcile(param.value)
         return True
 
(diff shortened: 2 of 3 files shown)


jpremill:
@sastraxi last night I saw this issue and I swear it was fixed but now I'm not sure it even helps it. Closing this for now. I deployed last night to my device and it did work... odd

sastraxi:
Hmm. The non-echoing sinks shouldn’t need reconcile, they’re meant to be commit only iirc. If you can provide repro steps (make sure you’re on the latest OTA MOD-UI as well as pi-stomp via Updates and Recovery) happy to jam on a solution together

jpremill:
@sastraxi  I just had this happen the other day again at rehearsal. this is a fresh 3.x install whatever the latest was at the time. the way I repro'd it was to turn on the device and then use long press to toggle to the next preset. once that was done it seemed to be a problem

jpremill:
turns out it was one of the pedalboards that caused this behavior.  I have no idea how or why but one set of pedalboards just flat out will not let me change any values.

jpremill:
Was not just the pedalboard. The only way to update a value is to use the first encoder. The value display overlay flat out doesn't work.

PR #240, opened 2026-08-23
Introduce a metronome when transport is rolling

sastraxi:
* fixed-config metronome that follows BPM / beats-per-bar
* sample-accurate metronome audio, pre-rendered samples in a python JACK client
* the metronome audio is combined with the pedalboard output and is pre-master gain
* Metronome ON/OFF toggle in the Audio & MIDI menu, replacing VU calibration
* VU calibration moved to the button bar on Audio & MIDI, replacing Restart (sound engine).
* Audo & MIDI becomes a metronome icon when it's enabled and transport is rolling

### Author checklist
* [ ] add volume control


rreichenbach (approved):
Wow, that's cool.  Love the new audio page.  I know there's not much room, but a separate volume control for the metronome would sure be nice.

PR #241, opened 2026-08-24
Fix AudioMidiPanel missing _reset_to_default

sastraxi:
ArcKnobWidget fires _reset_to_default on LONG_CLICK, but AudioMidiPanel lacked this method, causing AttributeError.

This fix matches the pattern used by mixer/panel.py.


rreichenbach (approved):
Looks good.

PR #244, opened 2026-08-29
Disabled menu items for read-only USB drives (fixes crash)

sastraxi:
We cannot backup onto read-only USB drives, but we assume all USB drives are read-write. This branch fixes that; read-only drives can still be selected as restore sources, but only if they truly have been backed up to previously.

I discovered this crash when I tried to Restore while my Launchkey Mini mkii was plugged in; this USB controller contains an internal hub that provides both a USB-MIDI device as well as a read-only flash drive with manual / driver installer.


rreichenbach (approved):
Looks good.

PR #245, opened 2026-08-29
[WIP] ADC toggler

sastraxi:
Like the footswitch bar, but for enabling/disabling the expression pedal. I really don't know if I like this.


sastraxi:
@rreichenbach I'm curious to get your take here at some point. The problem I'm trying to solve: wouldn't it be nice if users could get their expression pedal working without having to edit a file?

PR #246, opened 2026-08-29
Fast black screen shutdown and remove LCD splash

sastraxi:
* remove the built-in splash now that `lcd-splash` takes care of it for us
* remove `has_system_splash` (always True now)
* ensure that subprocesses get joined quickly so that they don't block shutdown
* add a `dismissable` arg, True by default, to message dialogs
* uses non-dismissable modal dialogs for terminal states (Shutting down..., Rebooting..., Restarting Sound Engine...)


rreichenbach (approved):
I think I kindof preferred the screen transitions (restart sound engine, shutdown) without the black screen, but this if fine too.  And yeah, good to get rid of the old splash.

sastraxi:
That’s fair—maybe a better approach would be to put up a standard “Please
wait…” modal rather than blacking out.

PR #248, opened 2026-08-31
Add back the state files

sastraxi:
Or: how I learned to stop worrying and love Chesterton's Fence

These state files are only resored if the current alsa global state doesn't contain the current sound card's name, so this doesn't obliterate user settings.


rreichenbach (approved):
thanks.

PR #249, opened 2026-09-01
Show disabled EQ for non-iqaudio cards

sastraxi:
In the Audio & MIDI menu:

* DISABLED -> N/A for badge text
* Show equalizer as disabled for non-iqaudiocodec cards (in addition to unsupported sample rates)


rreichenbach (approved):
looks good.

PR #250, opened 2026-09-01
Scroll pedalboard/snapshot titles only when selected

### LINKED ISSUE #46: Periodic clicking noise at high gain on pi-Stomp OS version 3.3.0
I'm experiencing a strange periodic noise on my pi-Stomp v3 running pi-Stomp OS 3.3.0.

The noise sounds like a regular "click/tick", almost like a metronome. It becomes noticeable when the gain is set high.

What I've observed

- The noise occurs even with no guitar connected.
- It happens with a completely empty pedalboard, with no effects loaded.
- It is not caused by my guitar or instrument cable.
- I'm using the official Raspberry Pi power supply.
- The noise appeared after updating to pi-Stomp OS 3.3.0.
- I initially reinstalled the same 3.3.0 version, and the noise disappeared temporarily. The problem returned shortly afterwards, when I created a pedalboard (I'm not sure whether creating the pedalboard caused it or whether it was just a coincidence).
- Interestingly, when booting into recovery mode, the noise disappears completely.

Reproduction

1. Boot pi-Stomp v3 with OS 3.3.0.
2. Disconnect the guitar from the input.
3. Load a high-gain NAM amp model, or use a clean/empty pedalboard and increase the gain sufficiently.
4. Listen to the output.

The result is a periodic clicking/ticking noise, similar to a metronome.

Expected behavior

With no input signal, the output should remain quiet apart from the normal noise floor, even at high gain.

Actual behavior

A periodic clicking/ticking noise becomes audible at high gain, despite there being no input signal.

Interestingly, the noise disappears completely when booting into recovery mode.

I'm not sure whether this could be related to ALSA/JACK, the audio codec/mixer configuration, or another change introduced in 3.3.0

Any help identifying the cause would be appreciated.

sastraxi:
SPI transfers cause small bursts of noise on ADC inputs. These are usually completely inaudible, but some high-gain amplifiers and NAM models can make these transfers audible. This is normally fine: you're unlikely to be interacting with the screen when playing. However, it does mean that any animations on the screen can cause persistent noise.

This PR makes it so that these widgets only scroll when they're highlighted, giving users a way to avoid the noise if it's affecting them.

Potentially closes https://github.com/TreeFallSound/pi-gen-pistomp/issues/46

### Author checklist

- [x] re-write llm slop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Notable user visible changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- Pedalboard/snapshot titles now auto-scroll only while selected with the NAV encoder (at most one thing scrolls at a time, and they sit at their leftmost position otherwise) — LCD updates over SPI are audible on the DAC at high gain, so the screen stays quiet while you play
+### Fixed
+- Long titles scrolled slightly past their last pixel before bouncing back; the scroll window now matches the drawn text area exactly
+
 ## [v3.3.1] - 2026-08-14
 ### Fixed
 - Pedalboards list no longer shows deleted pedalboards after they are removed from MOD-UI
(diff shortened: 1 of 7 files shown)


rreichenbach (approved):
looks good.

PR #251, opened 2026-09-05
Ensure that parameter commits happen (loading end)

sastraxi:
> Written mostly by Claude Code

Sometimes when users turned on their pi-Stomp, longpress worked to open a plugin's controls, but values refused to change. Other times, everything worked as expected. Knobs and encoders bound to plugin parameters continued to work just fine.

### Cause 1 — the suppression window stayed open

poll_modui_changes opened _is_pedalboard_loading at each change of last.json. Only a change of bundle closed it again, through set_current_pedalboard. If last.json named the bundle that was already current, no branch closed the window. _publish_plugin_param then refused all sends, and commit put the last confirmed value back on the screen.

At start this is a race. FileChangeMonitor records the mtime of last.json when Modhandler is constructed. MOD-UI then writes last.json during its own start, with the same bundle. If that write comes after the baseline, the first poll sees a change of the same bundle, and the window stays open. A save in MOD-UI does the same.

**Correction.** Only loading_start opens the window, and loading_end closes it. set_current_pedalboard also closes it, for a load that MOD-UI stopped before loading_end.

### Cause 2 — two dispatch paths for one bypass

The panel bypass button wrote the local value and sent its own WebSocket message, so a bound footswitch did not follow it. toggle_plugin_bypass held a second fork: for a plugin with a footswitch, it made a false press. The binding table then fired the row that won that switch, and that row is not always the bypass.

**Correction.** Each bypass from the UI is one commit on the :bypass parameter. _sink_for selects the transport — the MIDI CC of the bound footswitch, or the WebSocket. The keycap follows, because StatefulController subscribes to the settled value. The UI never makes a false press.

### Cause 3 — a lost edit under "back-pressure"

_flush_param_queue ignored the result of the send and emptied the queue, so the LCD kept a value that MOD-UI did not have. Backpressure (too many items in the send queue) makes a value late, not incorrect. A refused send now stays in the queue and goes again at the next tick, and a newer value for the same symbol replaces it.

I propose that we remove the back-pressure algorithm completely, and instead invest in logging that will tell us if   MOD is not able to keep up with our messages (unlikely outside of blend mode).   

### A defect that we made, and found, in this work

The first correction to cause 2 sent each footswitch-bound parameter out as MIDI CC. A footswitch CC has two codes only. MOD-UI can learn a footswitch onto a continuous parameter, with a minimum and a maximum from its advanced menu. A press moves correctly between those two values. But an edit from the UI that stopped between them became an endpoint, so the screen showed one value and mod-host held a different one.

_publish_switch_cc now sends the two ends of the binding range as CC, and all other values on the WebSocket. Three tests hold this. No user saw the defect.

### Why one control answered for two parameters

A binding is a CC number, and a CC number is an address, not a name. Two parameters can name one address. The old code found the control from the address, then had to ask a second time if that control was really the correct one. Current now keeps the decision of the binder, and control_for reads it. One place resolves the address.

### Removals

* has_footswitch had no reader after the dispatch fork went away, so the field, its two writers and 60 tests are gone
* BypassSource did not earn its keep
* The ws_bridge is not None guards are gone, because the property already asserts that the bridge is there

### Not covered

PluginPanel._send_param writes to ws_bridge directly and does not use _sink_for. So any panel edit ignores the suppression window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- The parameter dialog on the LCD sometimes did not change values due to a race condition with respect to MOD-UI's `last.json`. pi-Stomp then did not send parameter changes to MOD-UI until you selected a different pedalboard. Parameters on a knob or an encoder continued to work, because they send MIDI CC.
+- The LCD showed a bypass that MOD-UI did not receive, if you tapped it while a pedalboard loaded. The LCD now keeps the last value that MOD-UI confirmed.
+- A parameter change from a plugin menu could be lost if the connection to MOD-UI was busy. pi-Stomp now sends the value again on the next cycle.
 
 ## [v3.3.1] - 2026-09-01
 ### Added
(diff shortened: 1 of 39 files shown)


rreichenbach (approved):
This seems to be working nicely as expected now.  I hadn't realized/appreciated how complex the param change sequence was.

jpremill:
Param changes definitely felt complex this sounds like it might address the issue I was having.

sastraxi:
Yeah, I'm not super satisfied with how this has all turned out. Hopefully this is a step in the right direction.

PR #257, opened 2026-09-08
Re-render param badges in open menus on MIDI learn

sastraxi:
Previously, when you opened up a plugin menu it would properly show badges i.e. (1/2/A/B/C/D) on bound parameter editors, but these badges did not update when changes are made e.g. via MIDI learn, or removing bindings in MOD-UI.

This PR fixes that by creating a binding "revision" that monotonically increases whenever a binding is updated. Each UI tick, plugin menus now compare the current revision to the current one being advertised; any difference wins (to account for overflow). Badges are then re-calculated and re-drawn, if required.

### DIFF
--- a/modalapi/modhandler.py
+++ b/modalapi/modhandler.py
@@ -1483,6 +1483,10 @@ class Modhandler(Handler):
         render from."""
         return self._controller_manager.effective_table
 
+    @property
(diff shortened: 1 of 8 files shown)


rreichenbach (approved):
This seems to work with everything I tried.

PR #261, opened 2026-09-18
Remove audio card seeding and state files

sastraxi:
The interactive `change-audio-card.sh` script lives on, delegating its responsibility to [pistomp-audio 1.1.0+](https://github.com/TreeFallSound/pi-gen-pistomp/pull/58)'s `seed.sh`. As such, the released version in pi-gen-pistomp must declare it as a dependency.


rreichenbach (approved):
looks good.

